Issue #245, opened 2014-09-10
multiple collections / database support?


hayksaakian:
I've run into performance issues storing massive (1gb+) amounts of data in one single indexeddb in the past with brian.io/lawnchair.

I solved this with multiple distinct databases for each logical collection.

I request that localForage support using multiple instances and databases simultaneously.

I might implement this myself if lawnchair stops being good enough.

Issue #45, opened 2014-02-12
Bake in namespacing


jwlrs:
It would be awesome for this library to add namespacing as a core concept from the start.  This would mean other libraries could be built that depend on localForage and do so in a way that avoids collisions/conflicts.  

See https://github.com/joelarson4/LSNS for my train of thought.

PR #253, opened 2014-09-25
feat(localforage): add support for multiple storage instances

thgreasi:
Related to #245, #45 and #4.

**Note**: the two new test cases fail localy on my machine, but pass on TravisCI!?!? The test cases also pass if run them on a browser.
**Update**: [here is a fiddle ](http://jsfiddle.net/e1bpzjgd/1/) with the running example.

### DIFF
--- a/src/drivers/indexeddb.js
+++ b/src/drivers/indexeddb.js
@@ -22,6 +22,7 @@
     // Open the IndexedDB database (automatically creates one if one didn't
     // previously exist), using any options set in the config.
     function _initStorage(options) {
+        var _this = this;
         var dbInfo = {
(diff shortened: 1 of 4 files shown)


gustavohenke:
Guys, I see `createInstance()` when I type `localforage` in the console, but it's not documented anywhere :)

gustavohenke:
Sorry, didn't see #260.

perlun:
See also #367. TL;DR - setting `storeName` doesn't seem to work for me w/ Chrome Canary. I suspect it is because `_initStorage()` does not get properly called. (more details in the referred-to issue)

PR #306, opened 2014-11-19
Localforage, browserify and IE8 do not play nicely together

jamessharp:
When requiring localforage with browserify IE8 baulks with 'Object doesn't support this property or method' type errors. 

This is due to the fact that browserify will bundle up the `promise` module which is not es3 compatible. The compiled version (in the dist directory) uses a different promise library - `es6-promise` which _is_ es3 compatible. This PR contains a tweak so that browserify will also bring in `es6-promise`.

It's not a very elegant solution, in fact it feels a bit hacky, but I erred on the side of caution and tried not to touch anything.

Note that even with this change a browserified version of this package is still not quite IE8 ready. For that you need to filter out the es3 reserved keywords using something like [dekeywordify](https://github.com/pluma/dekeywordify)

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,13 +37,17 @@
     "phantomjs": "^1.9.7-12",
     "uglify-js": "^2.3.x"
   },
-  "browser": "src/localforage.js",
+  "browser": {
+    "dist/localforage.js": "./src/localforage.js",
+    "promise": "./src/browser-promise.js"
+  },
   "main": "dist/localforage.js",
   "licence": "Apache-2.0",
   "bugs": {
     "url": "http://github.com/mozilla/localForage/issues"
   },
   "dependencies": {
-    "promise": "^5.0.0"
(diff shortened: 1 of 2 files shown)


thgreasi:
This might be obsolete, since in v1.3 (source found in master) we will not be requiring the promise polyfill manually (since it's an anti-pattern for many) but instead bundle it first and expect to be available.

tofumatt:
I think yes. I'll say it's related to #432 so I'm reminded to check after we merge it in.




- tofumatt

thgreasi:
Just did a test on IE8 emulaton (within IE11) and everything worked fine with the dist build.

thgreasi:
But you are right, we should close this after v1.3 is out, since by then everybody will be pointed to the dist build.

tofumatt:
Closed in 690c73df5e2263b171a23e440d285a642cdc6af4. :smile:

PR #318, opened 2014-12-07
Bulk save functionality (setItems method) + other minor fixes

lu4:
No new tests, existing tests work

Issue #175, opened 2014-05-26
Strange behaviour of clear function callback


alexandremottet:
Hi,

I have notice a strange behaviour (maybe a bug) in the clear function.
There is a Codepen example : http://codepen.io/anon/pen/vpqzt?editors=101

```
localforage.setItem('hello', 'Hello World !', function(value) {
  logs.append("save hello key with value : " + value + "<br />");
  localforage.clear(function() {
    logs.append("Database is now empty.<br />");
    localforage.getItem("hello",function(data) {
      logs.append("getItem hello with value "+data+".<br />");
    });    
  });  
});
```

On Firefox, the callback is called properly and I received `null` value from `getItem`.
On the other hand, on Chrome (last version) when I get into the getItem callback the old stored value (`Hello World !`) is still present.

PR #320, opened 2014-12-09
Batch writes

### LINKED ISSUE #301: batch writes
While working on remotestorage.js I found that IndexedDB is very slow for writing if (like localForage) you use it as a key-value store. This is because each write is a transaction that can be rolled back. In remotestorage.js 0.10 we fixed this with a ["commit cache"](https://unhosted.org/practice/27/Persisting-data-in-browser-storage.html#persisting-client-side-data).

Batching writes is quite easy:
- keep track of the batch that is currently running, and of the write requests that are queueing up
- when the current batch finishes, you start a new batch, until nothing more is queueing
- when reading, first check the queued items, then the items in the running batch, then check IndexedDB

To the web developer who uses it it's all transparent, and in applications that store 100 or 1000 items, it's orders of magnitude faster. You can see an example here: https://michielbdejong.com/commit-cache.html

Would you like me to send a pull request for adding such a commit cache to localForage?

michielbdejong:
Batch writes together automatically, without changing the API. Fixes #301.

~~TODO: add tests~~
TODO: ~~find a way to stub indexedDB, so that I can~~ test the [`doTask` and `_updateDatabase` functions](https://github.com/michielbdejong/localForage/blob/batch-writes/src/drivers/indexeddb.js#L128-L179)

Sorry for adding so many private funtions on the asyncStorage object, let me know if there's a better way to do this.

I hope this is more or less what you were expecting, there are obviously other ways to architect the code. I'll be happy to change it if you point me to examples of what it should look like.

### DIFF
--- a/src/drivers/indexeddb.js
+++ b/src/drivers/indexeddb.js
@@ -50,6 +50,130 @@
         });
     }
 
+    function doTask(store, task) {
+        if (task.action === 'put') {
+            return store.put(task.value, task.key);
+        }
+
+        if (task.action === 'delete') {
+            // We use a Grunt task to make this safe for IE and some
+            // versions of Android (including those used by Cordova).
+            // Normally IE won't like `.delete()` and will insist on
+            // using `['delete']()`, but we have a build step that
+            // fixes this for us now.
+            return store.delete(task.key);
+        }
+
+        if (task.action === 'clear') {
+            return store.clear();
+        }
+    }
+
+    function _updateDatabase(tasks) {
+        var self = this;
+        var promise = new Promise(function(resolve, reject) {
+            self.ready().then(function() {
+                var dbInfo = self._dbInfo;
+                var transaction = dbInfo.db.transaction(dbInfo.storeName, 'readwrite');
+                var store = transaction.objectStore(dbInfo.storeName);
+
+                for (var i=0; i<tasks.length; i++) {
+                    tasks[i].req = doTask(store, tasks[i]);
+                }
+
+                transaction.oncomplete = function() {
+                    resolve();
+                };
+                transaction.onerror = function() {
+                    reject();
+                };
+
+                // The request will be aborted if we've exceeded our storage
+                // space. In this case, we will reject with a specific
+                // "QuotaExceededError".
+                transaction.onabort = function(event) {
+                    var error = event.target.error;
+                    if (error === 'QuotaExceededError') {
+                        reject(error);
+                    }
+                };
+            }).catch(reject);
+        });
+
+        return promise;
+    }
+
+    function _collectBatch() {
+        var batchSize = 0;
+        var  keyAlreadyAffected = {};
+
+        if (this._taskQueue.length === 0) {
+            return []; // Nothing to do
+        }
+
+        if (this._taskQueue[0].action === 'clear') {
+            return this._taskQueue.splice(0, 1); // Execute 'clear' operation alone by itself
+        }
+
+        while (true) {
+            if ( (batchSize === this._taskQueue.length) ||             // Batch will equal the entire queue
+                    (this._taskQueue[batchSize].action === 'clear') ||     // Batch will end before a 'clear' operation
+                    keyAlreadyAffected[this._taskQueue[batchSize].key] ) { // Batch will end before one same key is affected again
+                return this._taskQueue.splice(0, batchSize); // Execute tasks [0 .. batchSize-1]
+            }
+            keyAlreadyAffected[this._taskQueue[batchSize].key] = true;
+            batchSize++;
+        }
+    }
+
+    function _nextBatch() {
+        var self = this;
+        this._scheduled = false;
+        this.ready().then(function() {
+            var batch = self._collectBatch();
+            if (batch.length === 0) {
+                self._running = false;
+                return;
+            }
+            self._running = true;
+            self._updateDatabase(batch).then(function() {
+                for (var i=0; i<batch.length; i++) {
+                    batch[i].resolve(batch[i].value);
+                }
+                self._nextBatch();
+            }, function() {
+                for (var i=0; i<batch.length; i++) {
+                    batch[i].reject(batch[i].req.err);
+                }
+                self._nextBatch();
+            });
+        }).catch(function(err) {
+            //IndexedDB not ready, reject everything
+            var i;
+            for (i = 0; i < self._taskQueue.length; i++) {
+                self._taskQueue[i].reject(err);
+            }
+            self._taskQueue = [];
+        });
+    }
+
+    //`task` should be an object.
+    //`task.action` should be one of ['put', 'delete', 'clear']
+    //`task.key` should be set for 'put' and 'delete'
+    //`task.value` should be set for 'put'
+    //`task.resolve` should be a success handling function
+    //`task.reject` should be a failure handling function
+    function _queueTask(task) {
+        var self = this;
+        this._taskQueue.push(task);
+        if (!this._running && !this._scheduled) {
+            setTimeout(function() {
+                self._nextBatch();
+            }, 0);
+            this._scheduled = true;
+        }
+    }
+
     function getItem(key, callback) {
         var self = this;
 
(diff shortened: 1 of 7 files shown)


tofumatt:
So far this looks really good.

Honestly, I like this transparent queuing without changing the API, so I'd say yes you're on the right track, and I prefer this approach over #318 which seems to require using the bulk save calls to get the performance of batch writes. This branch would solve all the same problems that #318 aims to solve, right?

michielbdejong:
> This branch would solve all the same problems that #318 aims to solve, right?

afaik yes. I don't think many programmers want/need to control where the exact batch boundaries fall, so automatic batching should be easier to use for them. @lu4 can you confirm?

michielbdejong:
Rebased on current master.

test/test.taskbatching.js
```diff
@@ -0,0 +1,156 @@
+/* global beforeEach:true, describe:true, expect:true, it:true, Modernizr:true */
+describe('Task Batching', function() {
+    'use strict';
+
+    beforeEach(function(done) {
+        if (Modernizr.indexeddb) {
+            localforage.setDriver(localforage.INDEXEDDB, function() {
+                window._batches = [];  // FIXME: Is there a better place to store this stub recording?
```
tofumatt: Is this only stored for the test? If not, I would think just attaching it to the localForage instance would be best, either way. eg:

```javascript
var localforage = LocalForage.createInstance();
console.log(localforage._batch);
# => []
```

It shouldn't be a "public" variable or part of the API but I think that would work.

tofumatt:
This is looking really good. I like the minimal amount of code changing and the tests. :+1:

lu4:
Guys, its been a while when I was looking into GitHub, I've reviewed your question Micheil, and if you ask me, I would still go for two methods, both `setItem` and `setItems`. As I've already written in one of my previous responses, `setItem` with batching will pose significant memory overhead (in my case it's crucial) and by having only one setItem(with batching) method you are leaving no choice for people who may need to go around such inefficiency

thgreasi:
I think that setItems can be implemented as an extension of localForage.
Actually, it might even be possible to write batching as an extension, but
it would probably require more monkey patching.

I was actually thinking of mocking an extension with setItems for some time
now, but never got enough time.
Any thoughts?

tofumatt:
What are the memory/perf implications? The speed of the writes appeals to me here… would it make sense to offer batchWriting as a config option? Maybe even on by default?

michielbdejong:
I re-read @lu4's earlier remarks about memory efficiency [here](https://github.com/mozilla/localForage/issues/301#issuecomment-66114513) and [here](https://github.com/mozilla/localForage/issues/301#issuecomment-66153205), but still don't understand exactly what the argument is. Here's how I see it:

Memory usage is always equal to all the data that has already arrived in the browser (for instance, from a download or a file upload) but has not yet been committed to disk. So there are two situations:

* data trickles in, file by file (in this case automatic batching is optimal, and uses *less* memory than building up one big batch)
* all items are available at once (in this case both methods use the same amount of memory, even though explicit `setItems` does so for a shorter time period).

The only reason I can see to ever use explicit `setItems` instead of automatic batching would be that for saving, say, 1000 items it reduces the number of trips to disk to just 1, and automatic batching still needs 2 trips: first it saves the first item, and then it does a second trip to disk to save the other 999 items. (note to self: and maybe even that effect can be avoided by adding a `setTimeout(.., 0)` [here](https://github.com/michielbdejong/localForage/blob/batch-writes/src/drivers/indexeddb.js#L235), have to try that out).

test/test.taskbatching.js
michielbdejong: Thanks! moved it.

michielbdejong:
It looks like Travis is only running mocha-unit and not the saucelabs tests. I'll see if I can open a saucelabs account myself and try to run all tests that way. I mean, they're executing, but `Modernizr.indexeddb` is `false` in PhantomJS, so it skips all the actual testing work.

michielbdejong:
I think there is something wrong with how grunt calls sauce, because it fails in the same way for me locally if I just do `SAUCE_USERNAME=foo; SAUCE_ACCESS_KEY=bar; grunt saucelabs-mocha`, but if I go into `node_modules/grunt-saucelabs/tasks/saucelabs.js` and hack the values of `this.user` and `this.key` in a few places, it at least succeeds in setting up the tunnel to saucelabs.

michielbdejong:
Hm, "Disable Sauce Labs tests until we can fix them" https://github.com/mozilla/localForage/commit/17141c9ac47e327c98340cb26a7911fbd8683a6c seems related.

@tofumatt how can I run the new tests from this PR? Trying to view /test/test.main.html locally with a browser gives a blank page as well.

tofumatt:
I’ll have to check on this, I need to figure out why Sauce Labs isn’t working. I can test it locally soon though, but I’m probably going to be offline until tomorrow.

daseinhorn:
I'm really waiting for this to get merged. I ran into massive performance problems due to this and I would love to get the batch feature merged. Is there any progress to report?

tofumatt:
I've been moving countries and really backed up but I'll make an effort to
look at it today or tomorrow. Promise!

-tofumatt

daseinhorn:
@tofumatt thanks!

src/drivers/indexeddb.js
```diff
@@ -126,32 +126,145 @@
...
+                    }
+                };
+            }).catch(reject);
+        });
+
+        return promise;
+    }
+
+    function _collectBatch() {
+        var batchSize = 0,
```
tofumatt: Dunno why the linter didn't catch this, but each `var` should be on its own line.

michielbdejong: OK, fixed except [where it was already that way in master](https://github.com/mozilla/localForage/blob/master/src/utils/serializer.js#L163)

tofumatt:
Right now, looks like these tests are failing in Firefox. I need to fix up the sauce labs tests but I haven't been able to yet. You can see the tests locally by running:

```bash
grunt serve
open http://localhost:9999/test/test.main.html
```

michielbdejong:
OK, if I can at least see all the test output that way, that's already a big win - last time I tried I just saw a blank page. Will try to get the tests running+passing locally, at least for Firefox and Chromium on Ubuntu.

tofumatt:
Any news on this? If there's something you're specifically stuck or blocked on I should be able to help, finally found an apartment after moving so I'm less busy now :wink:

michielbdejong:
No, was just busy with other projects. Will try to find time this week!

tofumatt:
Sweet, thanks! :)

michielbdejong:
Fixed two tests, the ones still failing are:

* some of the ones I added (will look at those tomorrow)
* the four that also fail in master

tofumatt:
Which tests were failing in master? I notitced some were failing in asyncStorage, but I fixed them in #175. You might want to rebase on top of the new `master` branch.

michielbdejong:
@tofumatt can you think of more things to test? What other things could go wrong that are not covered in the new tests yet?

michielbdejong:
Rebase fixed those other four indeed, thanks.

tofumatt:
Now that everything is passing I'll take a look and see if we need anymore
tests, but the suite is pretty big already, so if batch writes are still
passing everything that's a great start. I'm getting on a flight but will
look at this more tomorrow.

-tofumatt

tofumatt:
I've looked through and I think this is all good. I'm gonna merge this in unless other folks have any thoughts. :smile:

tofumatt:
Actually, just ran through all the tests manually in Firefox and Chrome, looks like the `component` tests are failing [localhost:9999/test/test.component.html](http://localhost:9999/test/test.component.html).

Not sure why this would be as the require ones work fine. It's specifically Task Batching tests that are failing, and not even all of them.

tofumatt:
Especially strange: http://localhost:9999/test/test.component.html?grep=Task%20Batching shows them passing, http://localhost:9999/test/test.component.html does not.

michielbdejong:
maybe it's because of [timing](https://github.com/michielbdejong/localForage/blob/batch-writes/test/test.taskbatching.js#L4-L5) issues? How can I do that differently?

tofumatt:
I tried adjusting the time and giving it plenty of time to run and it still fails, but only with component. Very strange.

I'm looking into it a bit more, though if it's working with require and the regular JS file I'm tempted to say nevermind component unless a component user can tell me what to fix.

zawaideh:
Just curious if there is still a plan to merge this in?

tofumatt:
There is, but I have been woefully distracted. It’s my highest priority for the project though; looking to make some time later in the week.

zawaideh:
Thanks! And thank you for all the great work!

michielbdejong:
Great! Let me know if any action from my side is required.

ThomasConner:
Will this still be merged?

tofumatt:
I do plan to merge it but I need to spend some time tweaking it. If someone had the time to work on it I could tell them what needs to be done though.

- tofumatt (Sent from mobile)

ThomasConner:
What needs to be tweaked? I might have some time if you can point me in the right direction. Thanks for the great work!

tofumatt:
To start: **conflicts need to be resolved**. This patch is quite old and pre-dates our move to ES6 in our source files (#390).

That's the big one; let me look through other issues and report back in about thirty minutes :smile:

tofumatt:
From the looks of it that's the only issue with the code, though we'll want to manually test these in browsers to make sure it works in [all browsers localForage supports](https://github.com/mozilla/localForage/wiki/Supported-Browsers-Platforms) (sadly, we don't yet have sauce labs testing working to automate these tests, but I'd like that too).

If you resolve the conflicts and get it to where it could merge with `master` and pass at least the automated phantomjs tests that would be a **huge** help though. If you'd be willing to take that task on and push here once the tests pass/the code merges cleanly I can look again and see if there are browser-specific issues.

michielbdejong:
I can resolve the conflicts and put this feature "behind a flag". That would have the following advantages:

* it's not so scary, because most users will not enable it
* we can point users who suffer from slow writes to "try this experimental feature at your own risk"
* we can close this PR :) if people report it works well in practice, you can easily enable it by default

tofumatt:
That would be SO amazing, thank you a million. You are awesome, I owe you a coffee/beer/whatever :-)

michielbdejong:
Continued in #656

Issue #294, opened 2014-10-28
Pass iteration number to iterate callback?


tofumatt:
In the [API docs for `iterate`](https://github.com/mozilla/localForage/pull/283/files#diff-ecc2321a7154759fe07cea105cc32b08R321) we supply an example of early exit where the developer has to track the number of iterations, but should we just keep track of that ourselves and supply the number to the callback? Seems like something the developer shouldn't have to do manually.

Issue #281, opened 2014-10-19
Deprecate key() function


tofumatt:
(Discussion started in #280 and all the way back in #3.)

The `key()` function is certainly weird and I don't think 1:1 mapping of localStorage's API is the exact selling feature of localForage (even if it once was). In future versions deprecating this functionality might be smart to reduce the size of the library and the effort to write and test drivers.

I'm not sure what the actual gains to removing it would be, as it might not take up much actual code, but if no one seems to be using it we should drop support for it in future versions.

Thoughts?

PR #328, opened 2014-12-20
Fix for #294

psalaets:
Changes:
- 1-based iteration number is passed as 3rd param to iterator callback (#294)
- Removed the part about breaking iteration from the "should iterate [promise]" test. Breaking iteration is covered by another test.

I made it 1-based instead of 0-based because to me it answers "how many iterations so far?" as opposed to being like an array index. Is this violating principle of least surprise?

Maybe it should be 0-based so it can be passed into key()? Would that even work? Actually, seems like key() is falling out of favor (#281)


test/test.api.js
```diff
@@ -148,13 +148,20 @@ DRIVERS.forEach(function(driverName) {
...
 
-                            localforage.iterate(function(value, key) {
+                            localforage.iterate(function(value, key, iterationNumber) {
                                 accumulator[key] = value;
+                                iterationNumbers.push(iterationNumber);
                             }, function() {
-                                expect(accumulator.officeX).to.be('InitechX');
-                                expect(accumulator.officeY).to.be('InitechY');
-                                done();
+                                try {
```
psalaets: Sorry to bump a really old thread but my comment about why try/catch is necessary was wrong.

It finally dawned on me why these tests are timing out when an expect fails. Mocha is usually good at failing the test due to uncaught exceptions -- I just assumed it wasn't handling it in this case for some reason.

The reason why it's not happening here is because the Error never gets out of the Promise code. LocalForage invokes the callback inside a promise fulfill handler ([localstorage](https://github.com/mozilla/localForage/blob/master/src/drivers/localstorage.js#L296), [indexeddb](https://github.com/mozilla/localForage/blob/master/src/drivers/indexeddb.js#L378), [websql](https://github.com/mozilla/localForage/blob/master/src/drivers/websql.js#L384)), so if the callback throws (i.e. a failed expect) the Error is caught by the promise implementation which then "swallows" the Error and rejects the resulting promise.

try/catch in tests is still useful in this case but only because of how LF invokes callbacks.

PR #340, opened 2015-02-01
Test for commonjs first so webpack will work

k88hudson:
I'm seeing a problem consuming localforage in webpack right now that seems to be due to `define.amd` existing but webpack expecting commonjs-style exports. 

This is a working solution, but there might be a better/less fragile solution to this, happy to do something else if you have suggestions or I can find out how this is dealt with by other packages.

### DIFF
--- a/src/drivers/indexeddb.js
+++ b/src/drivers/indexeddb.js
@@ -401,12 +401,12 @@
         keys: keys
     };
 
-    if (typeof define === 'function' && define.amd) {
+    if (typeof module !== 'undefined' && module.exports) {
+        module.exports = asyncStorage;
+    } else if (typeof define === 'function' && define.amd) {
         define('asyncStorage', function() {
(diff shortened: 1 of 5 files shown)


tofumatt:
I think a lot of these things are a bit fragile, but frankly this looks fine to me. Still passes all the tests so until anyone else proposes anything fancier this is :+1: 

(Getting JS libs to play well with the countless module system confounds me as I don't have enough experience outside a few.)

tofumatt:
Thanks, PS!

tofumatt:
I'll actually publish a bugfix version for this, in case you're pulling this in via bower or npm.

tofumatt:
Tag pushed: https://github.com/mozilla/localForage/releases/tag/1.2.2

k88hudson:
Thanks, much appreciated!

PR #346, opened 2015-02-18
Fix docs inaccuracy re: setting localStorage key prefix

psalaets:
# The Problem

When localStorage is the underlying store, [the docs](http://mozilla.github.io/localForage/#config) say `storeName` will be the key prefix but [the tests](https://github.com/mozilla/localForage/blob/master/test/test.config.js#L200) check that `name` is the prefix.

# Change made

Copy/pasted the blurb about localStorage key prefixes from `storeName` to `name`.

### DIFF
--- a/site/index.md
+++ b/site/index.md
@@ -524,7 +524,8 @@ values can be set:
   <dt>name</dt>
   <dd>
     The name of the database. May appear during storage limit prompts.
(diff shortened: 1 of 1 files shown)


tofumatt:
Good catch, thanks! :+1:

PR #351, opened 2015-02-26
update jscs configuration to force 4 spaces indentation

stephanebachelier:
@tofumatt same things as in localforage backbone


tofumatt:
Excellent, thanks!

Issue #304, opened 2014-11-10
make `localforage.config` fail if `version` is not a number


lddubeau:
I set my `version` to a string value because I saw `1.0` in the [documentation](http://mozilla.github.io/localForage/#config) and just about every system I use that uses the dotted notation for version numbers treats version numbers as strings (because they typically support `major.minor.patch`). I did not notice the lack of quotes around it.

My fault, but it would be helpful to have `localforage.config` fail if the version number is a string. Why? IndexedDB in FF and Chrome will be happy with `"1.0"`, but IE's IndexedDB will throw `InvalidAccessError`.

Moreover, one of localForage's features is the ability to switch to a different driver if desired. I take it from reading the webSQL standard that it takes a string as a version number when opening a database. Someone who had code working fine with a version number set to a string value (e.g. "1.3a") because they used the webSQL or localStorage drivers will find their code failing if it runs with the IndexedDB driver.

PR #356, opened 2015-03-17
Fix for #304

irukavina:
Add check if the config() was called with not-a-number version
parameter and the corresponding test


tofumatt:
This is good but I think it would be even better to try to cast the string to a number first. So things like `'2.0'` become `2.0`. Otherwise I think we're effectively breaking the API (even if using it wrongly resulted in a different error).

Do you think we could try casting a string value first?

irukavina:
You're right - I'll modify the code and update the PR.

irukavina:
One thing that gives me concern is the following scenario: 

Imagine someone creating a WebSQL storage with localForage. One of the used options is version:
```
dbInfo.db = openDatabase(dbInfo.name, String(dbInfo.version),
                                         dbInfo.description, dbInfo.size);
```

If the casting fix is added, the version parameter would be first cast to a float, and then recast as a string. Thus, before the fix, if the version was 13.0a, openDatabase will be called like this:
```
openDatabase(_, '13.0a', _, _)  // no casting here
```
and after the proposed fix openDatabase would be called like this:
```
openDatabase(_, '13', _, _) // String(parseFloat('13.0a')) -> '13'
```

This call will result in an error since there is no db.changeVersion in the localForage's WebSQL driver: 
```
unable to open database, version mismatch, '13' does not match the currentVersion of '13.0a'
```

I think that either: 
- the check if float is not a number should be a bit smarter, throwing an error if there are characters that are not standard in a number (for example, alphanumerics) 

- a breaking change in the API will have to be added

Can you share your thoughts on this? 

Thanks

tofumatt:
Hmmm, I guess I'd rather the check be smart and only allow version strings that **all** drivers could use and didn't fail on.

If that requires a breaking change or even a version release with a deprecation then another with the breaking change I'm cool with that.

Issue #342, opened 2015-02-04
localforage.createInstance: cannot create several instances with the same name (=db name)


jlchereau:
```js
var table1 = localforage.createInstance({name: 'db', storeName: 'store1' });
var table2 = localforage.createInstance({name: 'db', storeName: 'store2' });
//Do something with table1 and table2
```
This fails on Windows 8.1 + Chrome 40 (with latest updates as of 04/02/2015).
The same with different db names ('db1' and 'db2') works but you then have one IndexedDB database per table.

![image](https://cloud.githubusercontent.com/assets/2556751/6039038/96075890-ac5a-11e4-8b8d-34a8faa425d1.png)

RobYed:
I can confirm, that createInstance with the IndexedDB driver currently doesn't create multiple instances with the same `name`. Whereas multiple instances with different `name` work. With the WebSQL driver this is not a problem at all. Same `name` but different `storeName` works well.

The feature itself is awesome! It's exactly what I needed in my app project :)

Tested with localforage version 1.2.2 on 
- Win8.1 + Chrome 43 canary
- Ubuntu 14.10 + Chrome 41
- Ubuntu 14.10 + Firefox 36 (IndexedDB driver)

perlun:
This is indeed a bug in `localForage`. See the referenced PR (#342).

sebweaver:
I've created the PR https://github.com/mozilla/localForage/pull/452 with a proposed implementation for fixing this issue. I hope it could help.

PR #367, opened 2015-03-31
Added documentation  for multiple instances

### LINKED ISSUE #260: Add API docs for new multiple storage feature
Since merging in #253, we should add some docs for it. :smile:

RangerMauve:
Added docs to both README and the site that mention the multiple instance functionality and provide an example. Fixes #260

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,6 +135,24 @@ localStorage make storing many large Blobs impossible.
 
 [api]: https://mozilla.github.io/localForage/#setitem
 
+## Multiple instances
+
+You can create multiple instances of localForage that point to different stores using `createInstance`
+
+``` javascript
+var store = localforage.createInstance({
+  storeName: "storeNameHere"
+});
+
+var otherStore = localforage.createInstance({
+  storeName: "otherStoreName"
+});
+
+// Setting the key on one of these doesn't affect the other.
+store.setItem("key", "value");
+otherStore.set("key", "value2");
+```
+
 ## Configuration
 
 You can set database information with the `config()` method.
--- a/site/index.md
+++ b/site/index.md
@@ -641,3 +641,25 @@ use this to make sure the browser in use supports your custom driver.
 </aside>
 
 [default drivers]: https://github.com/mozilla/localForage/tree/master/src/drivers
+
+# Multiple Instances
+
+You can create multiple instances of localForage that point to different stores.
+
+## createInstance
+
+This is used for creating a new instance of localForage
+
+``` javascript
+var store = localforage.createInstance({
+  storeName: "storeNameHere"
+});
+
+var otherStore = localforage.createInstance({
+  storeName: "otherStoreName"
+});
+
+// Setting the key on one of these doesn't affect the other.
+store.setItem("key", "value");
+otherStore.set("key", "value2");
+```


tofumatt:
Looks great! Not sure why Travis is complaining but I'll fix that then merge this in. Thanks  :+1:

RangerMauve:
Oh weird. It seemed to work yesterday.

tofumatt:
Yeah, I think it’s an issue with Travis or the config in it. It’s happened on other Pull Requests too but not locally. I’ll look into it :)

RangerMauve:
Sounds good! :D

psalaets:
The docs changes in this PR are different than what I've seen.

In my experience, only the `name` config matters for namespacing the data of localforage instances. I get errors when `name` is the same and `storeName` is different. Issue #342 may be an example of this.

As a sanity check I made an example of setting only `name`, setting only `storeName` and setting both: https://github.com/psalaets/localforage-multi-instance.

I could be missing something here, can someone double check this?

RangerMauve:
I based the docs around the [unit tests](https://github.com/thgreasi/localForage/commit/292c4e59b9c8a7b4a813b7b67fa64a2e0147a729) that were created when this feature was added in.

To be honest, I hadn't really tested it too much myself.

The tests only reference storeName, though.

psalaets:
Around line 399 at that link it sets `name` and `storeName`.

RangerMauve:
So it does. I must have missed that. I'll add it in.

tofumatt:
`name` should be considered the `database` -- in WebSQL it's the DB name and in IndexedDB it's the entire database identifier

`storeName` is a "table" or "individual store" name -- an entire app might go inside one `name` but with different model's data in each `storeName` eg like a MySQL Database with multiple tables.

Both options are available in the API in case you wanted to store everything in one DB with isolation via table/stores (this is when you would use `storeName`) or have multiple DBs via `name` to entirely encapsulate things (eg you had a library that stored data and used its own DB).

Does that make sense?

(In general, I'd say different stores in the same app might want to use the same `name` and different `storeName`s. I'm not sure what the performance implications of one over the other is though.)

psalaets:
That sounds ideal but I haven't been able to get that to work. Whenever I set only `storeName` like this

```js
var store1 = localforage.createInstance({
  storeName: 'store1'
});

var store2 = localforage.createInstance({
  storeName: 'store2'
});
```
I get errors in Chrome and FF.

Also, it looks like the localstorage driver only uses `name` when namespacing keys. See https://github.com/mozilla/localForage/blob/master/src/drivers/localstorage.js#L62

tofumatt:
Damn, that's not good.

Also, I know that newer versions have had issues with `createInstance`, see: #342.

perlun:
@psalaets - I see the same thing that you do. When I override the `name`, that takes effect as in overriding the IndexedDB database name (Chrome Canary). When I also try to set the storeName to the name of the "table" (instead of keyvaluepairs), I get these exceptions:

```
Uncaught (in promise) DOMException: Failed to execute 'transaction' on 'IDBDatabase': One of the specified object stores was not found.
    at Error (native)
    at http://localhost:4200/assets/vendor.js:74803:45
```

Using localforage 1.2.2, installed via Bower. I assume there is some hardwiredness to `keyvaluepairs` somewhere...

perlun:
Doesn't seem like it, so then it's probably a matter of things happening in an unexpected order or similar.

perlun:
The same error here, but no insight into the reason: http://stackoverflow.com/questions/26473502/cannot-create-second-objectstore-indexeddb

perlun:
This one is really fishy. I've checked the tests in `localForage` and they very much set the `storeName` to a custom value - and it works! Whenever I try to use it from my own app (setting up some fixture data in my QUnit `setup` handler), things go berserk and I get the exceptions mentioned above.

Ideas, anyone?

psalaets:
@perlun The multi-instance tests set `name` *and* `storeName` which works because setting `name` to unique values prevents the issue.

I'm referring to https://github.com/mozilla/localForage/blob/master/test/test.api.js#L639.

There the two instances are the default localforage instance which uses `{name: 'localforage', storeName: 'keyvaluepairs'}` and `localforage2` which uses `{name: 'storage2', storeName: 'storagename2'}`.

Are there other tests I'm forgetting that only set `storeName`?

perlun:
Thanks. You are right, it's when you try to use the same `name` but with different `storeName` you get this. I debugged this quite a bit yesterday - #374 has the details.

thgreasi:
Should we merge this? After a restart on travis the tests passed.
I would just suggest to change the `storeName` to just `name` since there still is the "multiple instances with same storeName" #342 issue.
We should also mention that you can pass all the `config` options.

thgreasi:
@RangerMauve can you do a rebase? In all my merge attempts your initial commit & authorship gets lost.

thgreasi:
@RangerMauve Just opened #442 that resolves conflicts and merges this.
No need to rebase yourself.
Any feedback is welcome.

RangerMauve:
Sorry for the late reply! I currently don't have stable access to the internet. Glad to see this merged in, though! :D

Issue #350, opened 2015-02-24
FYI: Web Workers Bug Is Resolved


leegee:
`README.md` says that Web Workers use is blocked by [Bug 701634](https://bugzilla.mozilla.org/show_bug.cgi?id=701634), but that seems to now be resolved/fixed.

PR #371, opened 2015-04-17
Updated web workers section

lewiscowper:
https://bugzilla.mozilla.org/show_bug.cgi?id=701634 is now closed

### DIFF
--- a/README.md
+++ b/README.md
@@ -173,8 +173,7 @@ define(['localforage'], function(localforage) {
 
 ## Web Workers
 
-Web Worker support in Firefox is blocked by [bug 701634][]. Until it is fixed,
-web workers are not officially supported by localForage.
+As of Firefox version 37, the IndexedDB API is now supported in Web Workers, which allows localForage to be accessed from them.
 
 [bug 701634]: https://bugzilla.mozilla.org/show_bug.cgi?id=701634
 


tofumatt:
I don't think this is true. We have web worker tests that are marked as "skip" right now, and running them from Firefox 37 outputs:

```
ReferenceError: window is not defined (http://localhost:9999/dist/localforage.js:919)
process.on/global.onerror@http://localhost:9999/bower_components/mocha/mocha.js:5736:10
```

I believe this is a straightforward fix, but we can't claim to work with web workers yet :cry:

tofumatt:
Additionally, these tests fail from Chrome for IndexedDB and WebSQL.

lewiscowper:
Ah okay, I just noticed that the bug had been marked as resolved,
everything seemed in order.

Is there a bug for the failing tests?

tofumatt:
Yeah, sadly we use window in some places we shouldn’t. But I think it’s an easy bug to fix… worker support with a caveat is better than none at all, so I’ll leave this open and merge it in once everything is fixed.

tofumatt:
Related to #350.

tofumatt:
We fixed this elsewhere, so closing this, but thanks for your contributions!

PR #373, opened 2015-04-24
refactor(localforage): unify the handling of the driver import code

thgreasi:
* It reduces the number of `return`'s, `resolve`'s and `self._extend`'s in the code.
* Uses a promise to abstract the driver import and unify how loaded driver object is handled.
* Follows the same pattern used by localstorage and websql drivers to import the serializer.


thgreasi:
It should also make it easier for us to introduce more module loaders.
Moreover, we will need less refactoring in case (at some point) we move to es6 modules with a programmatic import API (lets say system.import for example).

tofumatt:
Oh, much nicer! :+1:

PR #374, opened 2015-04-28
test(createInstance): add test for same-name multiple instances #342

perlun:
We have a bug at the moment when you try to use the same `name` but with different `storeName` values. The exception you get in Chrome is `Failed to execute 'transaction' on 'IDBDatabase': One of the specified object stores was not found.`.

I've digged quite a bit into this and the problem is in this are: https://github.com/mozilla/localForage/blob/master/src/drivers/indexeddb.js#L41-L48

The `createObjectStore` is **only** possible to run from within an `onupgradeneeded` event. If you try to run it elsewhere, you get an exception (it needs the "version upgrade" transaction to be active).

Now, the thing here is that you only get this event when the *version* changes for the database (naturally...). This is how IndexedDB works, AFAIK, and I don't know any way around it. This means that `onupgradeneeded` will only get called for the very first `storeName` you provide; if you then try to create another `localForage` instance with another `storeName`, you will not get an `onupgradeneeded` event at all.

Possible workarounds:

1. Create the database differently, perhaps by having a list of `storeNames` in the options to `localForage.createInstance`. Loop over this list of `storeNames` in the `onupgradeneeded` event handler.
2. (Ab)use the version number handling in IndexedDB. I.e. increment the version every time we open the database... This is doable but requires a bit of quirks to get it working. This would probably provide the "cleanest" API to the user (but will be a bit of fiddling in the code with opening DB, querying version, closing, reopening with different versions etc...)
3. Maybe others. :smile: 

I think 1 is easiest and will probably implement that later during the day (European time). If anyone has better ideas, now is the time to suggest! :smile:

### DIFF
--- a/test/test.api.js
+++ b/test/test.api.js
@@ -634,6 +634,7 @@ DRIVERS.forEach(function(driverName) {
     describe(driverName + ' driver multiple instances', function() {
         'use strict';
         var localforage2 = null;
+        var localforage3 = null;
         var Promise;
 
         before(function(done) {
@@ -647,6 +648,17 @@ DRIVERS.forEach(function(driverName) {
                 // TravisCI seem to work fine though.
                 size: 1024,
                 storeName: 'storagename2'
+            });
+
+            // Same name, but different storeName since this has been malfunctioning before w/ IndexedDB.
+            localforage3 = localforage.createInstance({
+                name: 'storage2',
+                // We need a small value here
+                // otherwise local PhantomJS test
+                // will fail with SECURITY_ERR.
+                // TravisCI seem to work fine though.
+                size: 1024,
+                storeName: 'storagename3'
             });
 
             Promise.all([
@@ -690,7 +702,8 @@ DRIVERS.forEach(function(driverName) {
         it('retrieves the proper value when using the same key with other instances', function(done) {
             Promise.all([
                 localforage.setItem('key', 'value1'),
-                localforage2.setItem('key', 'value2')
+                localforage2.setItem('key', 'value2'),
+                localforage3.setItem('key', 'value3')
             ]).then(function() {
                 return Promise.all([
                     localforage.getItem('key').then(function(value) {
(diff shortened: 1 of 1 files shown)


tofumatt:
Any news on a patch to this or should someone else (me? :wink:) take a look at fixing it?

thgreasi:
I might be missing something but, shouldn't this test actually fail?

tofumatt:
I think it only fails on Chrome, which right now Travis doesn’t test. It should fail if you open the test page in Chrome though.




This really just reminds me I need to get sauce labs working with local forage.


- tofumatt

zcsongor:
Any news?

tofumatt:
Sorry, I haven't been able to look at this. I'm pretty busy this week with some other Mozilla work but I can review any pull request that fixes this ASAP. Writing the tests might take me a bit more time.



- tofumatt

test/test.api.js
```diff
@@ -649,6 +650,17 @@ DRIVERS.forEach(function(driverName) {
...
+                // otherwise local PhantomJS test
+                // will fail with SECURITY_ERR.
+                // TravisCI seem to work fine though.
+                size: 1024,
+                storeName: 'storagename3'
+            });
+
             Promise.all([
                 localforage.setDriver(driverName),
                 localforage2.setDriver(driverName)
```
sebweaver: It lacks the following:

```javascript
localforage3.setDriver(driverName)
```
I think that's why tests pass whereas they should fail, depending on the driver used by default.

sebweaver:
On PR https://github.com/mozilla/localForage/pull/452 I've fixed the test as commented above, and proposed an implementation in order to make it pass again. This implementation covers the IndexedDB driver of course, but since the test also highlighted an issue in the localStorage driver, the implementation covers the latter too.

Perlun's analysis of the issue in IndexedDB is very accurate. However, I dismissed the first workaround because whereas it could be simple in appearance, it also reveals some pitfalls. First of all, it changes the current specification for creating a localForage instance. This could lead to some confusion when used with drivers which didn't need that trick in the first place. The second pitfall is the way to address the correct store, from a single localForage instance, without changing the current specification and implementation of other APIs such as `setItem` and `getItem`.

So I finally decided to base my implementation on the second workaround, except it doesn't increment the version every time the database is opened, but only when necessary, i.e. when a new store is detected.

perlun:
Awesome @sebweaver! Thanks a lot for fixing. I never actually did the fix myself in the end, but it's great that someone else (= you) got around to doing it.

Issue #241, opened 2014-09-01
Chrome 37 IndexedDB stores incorrect content type for blobs


nolanlawson:
Hello from PouchDB. :)

Just a heads up, in case you weren't aware of this bug: https://github.com/pouchdb/pouchdb/issues/2701 and https://code.google.com/p/chromium/issues/detail?id=408120.

From Chrome 38+, blobs will be stored just fine, exactly like IE and Firefox. In the meantime, 37 is out there in the wild, and it's got this bug. If you'd like to detect blob support, [here's the code we use](https://github.com/pouchdb/pouchdb/blob/eae1b9da321c6bdb91ec0f79bdc33487a7555117/lib/adapters/idb.js#L1239-L1274), which should work on Firefox, IE, and Chrome 36-38.

sheppard:
I came across the lack of Blob support in my own testing (see #380) on an Android 4.4.4 Samsung device (UA string claims "Chrome/28").  I worked around it in my vendored copy of localForage with a [UA sniff](https://github.com/wq/wq.app/commit/db26e665ef2b5ea77d1b9f179bab38692b3e8f6a#diff-f23e060863b5229048e4640b06ad0bb6R2162) marking `IndexedDB` as unsupported in any Chrome less than 38.  This effectively makes `localForage` use the `WebSQL` backend, which serializes blobs for storage.  I suppose an alternative would be to enable the serializer for the `IndexedDB` driver in this case (which would be effectively the same as to the PouchDB approach).  

I didn't submit a PR as I'm not the UA sniff is the right approach (better approaches have already been listed), and I assume not everyone finds `Blob` support mission-critical as I do.

nolanlawson:
@sheppard Yeah sorry, I did not follow up with the localForage fix after reporting this. But basically our "blob support" check looks like this now: https://github.com/pouchdb/pouchdb/blob/1093c845bac9550baf8b0394d59fed3e14efe185/lib/adapters/idb/idb-blob-support.js

Reason being, there were actually more bugs that popped up since v38, and blobs are only reeeeaaally supported in v43 (see [state of binary data in the browser](https://github.com/nolanlawson/state-of-binary-data-in-the-browser#readme) for the gory details).

nolanlawson:
Found some time tonight, so here we go: https://github.com/mozilla/localForage/pull/382 :smiley:

[...]

PR #382, opened 2015-05-25
(#241) - fix blob support in old Chrome IDB

nolanlawson:
Fixes blob storage in Android 4.4 and Chrome <43.
It borrows a lot of code from PouchDB to accomplish this.
Sorry for not using the serializer, but I didn't notice
it until I had already written the code.

One unresolved issue is that the LocalStorage adapter
still isn't saving the blob type, which is why I left
that test commented out. If it were uncommented, though,
you'd see it passing in Android 4.4's and Chrome's IndexedDB.

### DIFF
--- a/src/drivers/indexeddb.js
+++ b/src/drivers/indexeddb.js
@@ -19,6 +19,159 @@
         return;
     }
 
+    var DETECT_BLOB_SUPPORT_STORE = 'local-forage-detect-blob-support';
+    var supportsBlobs;
+
+    // Abstracts constructing a Blob object, so it also works in older
+    // browsers that don't support the native Blob constructor. (i.e.
+    // old QtWebKit versions, at least).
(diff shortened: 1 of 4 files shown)


nolanlawson:
Nevermind, managed to fix the blob type issue for LocalStorage as well. That test is now uncommented, passes in old Chrome as well as PhantomJS.

My fix for the blob serializer for LocalStorage is a bit hacky, but I think it's required if we want to maintain backwards compatibility. See my changes to `serializer.js` and let me know if they jive for you.

nolanlawson:
Shoot, just realized I didn't fix https://github.com/mozilla/localForage/pull/381, though. Probably that `makeBlob` shim should go in both `serializer.js` and `indexeddb.js` (or I should figure out how you're requiring stuff; didn't quite get it tbh...).

PR #384, opened 2015-05-26
fix(examples): fix requirejs example

thgreasi:


thgreasi:
We could probably also delete the `./examples/require.js` file.

thgreasi:
Should we also include an example that requires each driver independently?
Possibly also add such a test case as well?

tofumatt:
I'll add this in but I don't think we need those examples. As discussed dynamic requiring isn't super important to me :smile:

PR #385, opened 2015-05-27
test(requirejs): add test for unbundled module loading

thgreasi:


test/runner.js
```diff
@@ -50,11 +50,27 @@ if (!Array.prototype.forEach) {
...
+            paths: {
+                localforage: '/src/localforage',
+                localforageSerializer: '/src/utils/serializer',
+                asyncStorage: '/src/drivers/indexeddb',
+                localStorageWrapper: '/src/drivers/localstorage',
+                webSQLStorage: '/src/drivers/websql',
+                Promise: '/bower_components/es6-promise/promise'
+            },
+            shim: {
+                localforage: ['Promise']
```
thgreasi: This could probably be omitted if change the way we require the Promise module.
Additionally, we could also first test `this.Promise` before attempting to require the module.

PR #387, opened 2015-05-31
test(browserify): add test for browserify bundling

thgreasi:


thgreasi:
I guess this was the only missing module type test.

I think we should first check for Promises before requiring the module. Something like:
```js
var Promise = this.Promise ||
    ((typeof module !== 'undefined' && module.exports && typeof require !== 'undefined') ?
     require('promise') :
     this.Promise);
```

I might be wrong (as a browserify noob), but I think that without that check a browserify bundle equivalent to `localforage.nopromises.js` can't be created.

tofumatt:
I'm okay with that; I'm tempted to do a cleanup and offer fewer downloadable versions anyway. It's a bit weird to keep track of anyway. Until anyone asks for it I won't worry about that :smile:

Issue #270, opened 2014-10-08
Change internal driver identifiers


tofumatt:
It would be nice if we didn't use obscure identifiers like `asyncStorage` or `localStorageWrapper` as the strings that identify each driver. Sure, we ask developers to use constants (like `localforage.INDEXEDDB`), but I'd still like it if the strings reported by, say, `localforage.driver()` were a bit nicer. Just a nice-to-have.

PR #390, opened 2015-06-06
refactor(localforage): use es6 with babel (loose all)

### LINKED ISSUE #158: Try writing some code using ES6
It might be nice to experiment with some ES6 goodness in localForage. And with #157, we can nicely transform it to ES5 for distribution.

thgreasi:
Resolves #158


thgreasi:
This PR should be initially be just for discussion until we discuss & strictly define things like,
where es5 sources should be placed (are they even needed?) etc...

Both refactoring commits pass all the tests.
The first one just uses `class` for `LocalForage` object and the second uses `export` statements.

Extra code needed for `Component` bundler (not covered by UMD) can be easily identified.
Still no webpack tests.

tofumatt:
:open_mouth: Soooo cool. I'm painting my apartment this weekend (and the router is in the room we're painting so I'm often offline :wink:) but I'll start review ASAP!

thgreasi:
Some things that need further discussion:


### remove `require('promise')`
For some developers requiring polyfills is an anti-pattern. Could we just add the polyfill to the dist bundle? Is this really required for a use case not covered by the test cases (nodejs maybe)?


### use es6 `import` statements
This can be used for the `serializer` module initially.
If we want to get rid of all the custom module imports then we need to:
* either drop support for dynamic driver loading with requirejs
* or implement a babel transformer for the not-yet-standardized `System.import`

This might also cause problems with component bundler!

thgreasi:
Also tried not to create a huge diff, in order to ease the reviewing process. We can later remove the outer `iife`s from all src files, since it is handled by the babel umd transformer.

thgreasi:
@tofumatt Just pushed [babel-plugin-system-import-transformer](https://github.com/thgreasi/babel-plugin-system-import-transformer) that replaces `System.import` to AMD, CommonnJS & Global module imports.

I have also created [a separate localforage branch](https://github.com/thgreasi/localForage/tree/es6-babel-system-import) that uses `System.import` statements. The only remaining `require` statements are those for the Promises polyfill for CommonJS (but we might remove it in the future).

Feel free to review. Should I push the commit in this PR?

tofumatt:
Yeah, I think pushing that commit to the PR makes sense. It simplifies the import system a lot, right?


- tofumatt

thgreasi:
I think so. I will push the commit tonight.

thgreasi:
Just pushed the `System.import` commit.

thgreasi:

tofumatt:
Sorry, I was off on the Mozilla work week :-/




I’ll look at it soon, though my friends are here from Montreal, so I might be delayed.


- tofumatt

Gruntfile.js
```diff
@@ -19,9 +19,48 @@ module.exports = exports = function(grunt) {
...
+                loose: 'all',
+                modules: 'umd',
+                moduleIds: true,
+                // sourceMap: true,
+                getModuleId: babelModuleIdProvider,
+                plugins: ['system-import-transformer']
+            },
+            dist: {
+                files: {
+                    'dist/es5src/localforage.js': 'src/localforage.js',
```
tofumatt: Why do we need to have an `es5src` folder of the individual components? I don't think this needs to be more than a build artifact. I would rather reduce the ways people can include `localforage.js`, so I don't mind if this means `require` people need to just import the whole file. I think that's easier to support.

dist/es5src/drivers/indexeddb.js
```diff
@@ -0,0 +1,576 @@
...
+    if (typeof define === 'function' && define.amd) {
+        define('asyncStorage', ['exports', 'module'], factory);
+    } else if (typeof exports !== 'undefined' && typeof module !== 'undefined') {
+        factory(exports, module);
+    } else {
+        var mod = {
+            exports: {}
+        };
+        factory(mod.exports, mod);
+        global.asyncStorage = mod.exports;
```
tofumatt: While we're changing so much about how things are defined and imported, it seems like a great opportunity to rename these drivers to something more meaningful (see #270).

tofumatt: Oops, sorry, this is all build artifact stuff, I'll move the comments.

tofumatt:
## re: `remove require('promise')`

I agree about requiring polyfills as anti pattern, at least in the `src` code. It's important to still include in the `dist` file, as [IE still needs the polyfill](http://caniuse.com/#search=promises). And I assume the polyfill is smart enough to back off from loading its own code if Promises are already available... for its file size I almost would rather just have one `dist` file as well. So loading promises into the `dist` file but otherwise not keeping them around is fine by me.

tofumatt:
### re: drop support for dynamic driver loading with requirejs

We should drop this support, so I'm fine with this.

Is there more `System.import` code you wanted to push here? Because even if we use that and it allows us to dynamically require drivers... I think that can be a hidden, unsupported feature (as it essentially is to me now). I don't see a lot of use in it (the drivers are reasonably small and the extra HTTP requests from not compiling them using a require.js compiler are likely as slow as the extra kB in size).

Gruntfile.js
thgreasi: How about just adding it to gitignore? Would this be enough?

dist/es5src/drivers/indexeddb.js
thgreasi: I think that it should be a separate PR. This one is already large enough to make tracking changes difficult.
Would be happy to open a new PR if you have suggested names on mind.
We just need to change:
* the driver names exposed by the localforage object
* the name property on each driver and
* the module mapping function inside grunt.js that babel uses

thgreasi:
> remove require('promise')

I will push a commit with this. It should be just a -4 diff.
We should be confident enough, now that we also have browserify tests. (Had some module loading environment isolation concerns.)

Gruntfile.js
tofumatt: Works for me, though I’d also be happy if it output to like a .tmp/ folder or something. Just to avoid confusion. :-)


- tofumatt

thgreasi:
> Is there more System.import code you wanted to push here?

No, it should be complete.

> drop support for dynamic driver loading with requirejs

Such change would allow us to:
* use plain es6 import statements on the top of the files and
* remove the babel plugin for system.import transpilation. (😫 this was the reason I wrote the plugin 😫)

If this going to be an unsupported feature, you might also want to remove the unbundled requirejs tests (or leave them as is or disable them to keep them as example code).

thgreasi:
At some point we could also remove the IIFEs from the source files on order to drop the loose babel flag. That's now required since es6 normally requires imports to appear on the top of each file before any code.

Did not include such a commit in order to make the reviewing process easier.

Gruntfile.js
thgreasi: How about `build/es5src` as a middle ground solution?

tofumatt: Sure, works for me. Just make sure it's in the `.gitignore` :smile:

tofumatt:
When I try to run these tests locally I get a Babel error:

```bash
~/Projects/localforage[thgreasi-es6-babel❗ ]% grunt test 
Running "babel:dist" (babel) task
Warning: Don't hotlink internal Babel files. Use --force to continue.

Aborted due to warnings.
```

tofumatt:
Yeah, re-running the Travis tests cause things to fail too: https://travis-ci.org/mozilla/localForage/builds/66916572 :disappointed:

dist/es5src/drivers/indexeddb.js
tofumatt: Fair enough. I can even do that in a separate PR after this. It's less important, you're right :smile:

thgreasi: It looks like they had a regression on v5.6.8. I will take a look at it
later today.

Gruntfile.js
thgreasi: `build` is already in `.gitignore` in the `### Ruby Stuff for slate docs` section.
`component.json` also needs references to the es5 sources. Removing them might cause problems to them.
Are we still gogo on this?

thgreasi:
It looks like they had a regression on v5.6.8. I will take a look at it later today.

thgreasi:
Just did a rebase and pushed the "remove promise polyfill require".
Also released a new version of babel-plugin-system-import-transformer to remove the warning.
The build seems to pass now.

tofumatt:
Cool, I’ll try to take a look tonight, but I have an early doctor’s appointment so we’ll see how far I get ;-)




If not tomorrow evening, thanks :-)


- tofumatt

thgreasi:
Just saw that, now that `require('promise')` is also gone, we no longer need all the manual module type detection code!
Pushed this as well.

tofumatt:
> Just saw that, now that require('promise') is also gone, we no longer need all the manual module type detection code!

Niiiiice.

tofumatt:
This looks pretty awesome. I can't see anymore issues, was there anything I didn't address or you think needs to be done?

I think it looks good to merge :+1:

thgreasi:
Hi Matt. I think that it's fine. I've been working on this since
February/March and I no longer see opportunities for fixes or improvements.
Thankfully we lately introduced more tests with the various module
packages/loaders.

This is a big enough PR already, so I guess that addressing webworkers and
changing driver names sound better be handled in separate PRs.

tofumatt:
Sounds good :-)




I'll merge it in today :-)




Thanks so much, it's awesome and I love how many lines of code you _removed_!



- tofumatt

thgreasi:
That puzzled me as well (the diff is +3,389 −3,685).
This might be virtual, possibly caused by the way babel or uglify formats
the code.

thgreasi:
Omg omg omg omg!
The time for es6 has come.

tofumatt:
:smile: :smile: :smile:

PR #400, opened 2015-07-06
Added two live demos to documentation

thrashergirl:
Demos are located in jsfiddle.


tofumatt:
(Oddly these don't work in Safari on OS X Yosemite, but I'm thinking that's either a Fiddle or a localforage bug.)

Looks great, thanks! :+1:

PR #401, opened 2015-07-08
Fixed typo in setItem() example

chrisledet:
Noticed a typo in one of the examples.


tofumatt:
Good catch! `image` is probably better than `img` though... any chance you could change both to the more verbose version? :smile:

chrisledet:
> Good catch! image is probably better than img though... any chance you could change both to the more verbose version?

No problem! This project is great by the way. Thanks!

tofumatt:
Thanks very much, and thanks for the contribution! :+1:

chrisledet:
Updated. Let me know if you need anything else.

PR #406, opened 2015-07-18
fix(localForage): support running inside web workers

### LINKED ISSUE #402: use localforage in a serviceworker
Hi,

I would like to use localforage in a serviceworker. Version 1.2.4 in Chrome 43 throws a reference error on line 963 "window undefined", because there is no window in the serviceworker. 

Window was removed in #144 but reintroduced in #237. 

Should localforage or should it not be usable in a webworker / serviceworker environment?

I'll see if I can get it to work.

thgreasi:
Related to #350.
Resolves #402.


test/test.webworkers.js
```diff
@@ -26,11 +26,17 @@ DRIVERS.forEach(function(driverName) {
...
         if (!Modernizr.webworkers) {
-            it.skip("doesn't have web worker support");
+            it.skip('doesn\'t have web worker support');
 
             return;
         }
 
+        if (driverName === localforage.LOCALSTORAGE ||
+            driverName === localforage.WEBSQL) {
+            it.skip(driverName + ' is not supported in web workers');
```
thgreasi: https://developer.mozilla.org/en-US/docs/Web/API/Worker/Functions_and_classes_available_to_workers

thgreasi:
@tofumatt , thanks to @remkoboschker 's help, this PR now also contains tests for service workers.
They currently do not run inside PhantomJS, only locally with [Chrome 44+ or firefox nightly with the dom.serviceworker enabled](https://github.com/mozilla/localForage/issues/402#issuecomment-124472099) (as @remkoboschker noted).
W8ting 4 review.

tofumatt:
Cool! I'm in Iceland for the week so I might not get to it until Sunday evening or Monday. But I'm on it ASAP!



- tofumatt

thgreasi:
I guess that Iceland-cool should be way cooler than Greece-cool, especially
this time of the year 😀

thgreasi:

tofumatt:
Sorry not yet but I'll check in a few hours. Iceland messed up my brain and I slept all afternoon! o_O



- tofumatt

dist/localforage.js
```diff
@@ -1828,6 +1829,10 @@ requireModule('promise/polyfill').polyfill();
         global.webSQLStorage = mod.exports;
     }
 })(this, function (exports, module) {
+    'use strict';
+
+    var _systemImportTransformerGlobalIdentifier = typeof window !== 'undefined' ? window : self;
```
tofumatt: Say that five times fast! :wink:

tofumatt:
Merged in cf6f1d11218e3453463f243e684d996c7a2eaf76!

thgreasi:
Tadaaa.
Late sleepless hot summer night's commits...
Can't even say it right two times.

dist/localforage.js
thgreasi: Sorry for the naming :disappointed: but this variable name is generated by babel-plugin-system-import-transformer@1.1.4 using the babel-provided `generateUidIdentifier` method to avoid conflicts.
The variable itself is used in the code transformations of `System.import` calls instead of just using `var global = window`.
This
* is required for web/service workers &
* allows for better minification, since it appears only once in the outemost scope of each transformed file that uses `System.import`. (In older versions this would appear once per `System.import` transformation.

It seems that [this was reverted](https://github.com/mozilla/localForage/blob/cf6f1d11218e3453463f243e684d996c7a2eaf76/dist/localforage.js#L1568-L1581) by ceab4ad15695e148e3a49ff2a3881264e3eeb841.

Can you please confirm that the latest versions of the npm package was used before this commit?

tofumatt: Oh shoot. I didn't mean to do that it must have done the build with an old version. I'll patch that up ASAP. Thanks. 



—

tofumatt: Should be fixed here: 0876fbcb5520d857d7d00b96e9a9edc81b8a76a6

thgreasi: Great! :+1:

lilred:
It'd be nice if support for web workers was documented :)

tofumatt:
@lilred If you could file a separate bug for that it would be great. Sadly support for web workers is spotty because Safari doesn't yet support everything we need to use IndexedDB with it, so only Chrome and Firefox support workers, if I recall correctly.

nolanlawson:
Safari now supports IDB in Web Workers [as of Safari 10](https://developer.apple.com/library/prerelease/content/releasenotes/General/WhatsNewInSafari/Articles/Safari_10_0.html). In Pouch, we've been evaluating the new WebKit IDB and it looks pretty good, although quite a bit slower than WebSQL. Needs more investigation though. I'm still waiting for someone to write the big "Is Safari IndexedDB ready yet?" blog post.

tofumatt:
Good to know! If there's a particular issue or somewhere I can follow along (or something I can do to help) let me know.

Issue #413, opened 2015-08-03
v1.2.6, Browserify and Ampersand-localforage 


rejas:
Since this weekend our build system chokes on the updated version of localforage. Using Browserify and ampersand-localforage we get this error:

13:31:18 [13:31:30] Browserify Error { [SyntaxError: Unexpected token (144:10) while parsing /home/.../node_modules/ampersand-sync-localforage/node_modules/localforage/src/localforage.js while parsing file: /home/.../node_modules/ampersand-sync-localforage/node_modules/localforage/src/localforage.js]

The version before just worked fine with the current ampersand-localforage/browserify :-(  Any hints on what we can change maybe on our side?

tofumatt:
No, this is my fault, sorry :cry: 

When we did the refactor into es6 we removed support for including `src/localforage.js` and require people to include the fully built version in `dist`. I've never officially supported including the source files but I think it's always worked until now and a lot of people have been doing it--but it causes us headaches :smile: 

I should have marked this as a version bump to 1.3 and didn't. I'll push 1.2.7 soon and it should continue working. Then I'll push 1.3, which won't work as you've been using but should work with the `dist/` version.

Sorry about all this. Should be fixed soon.

thgreasi:
Me sorry too :(

tofumatt:
Uploaded a new release to GitHub and npm. This should be fixed now, but there will be a new release later today: 1.3 that **will not support loading `src/localforage`** and will only work with the `dist/localforage.js` version.

thgreasi:
We  also have to change the "browser" field in package.json to target the
dist file as well.

On Mon, Aug 3, 2015, 14:57 Matthew Riley MacPherson <
notifications@github.com> wrote:

> No, this is my fault, sorry [image: :cry:]
>
> When we did the refactor into es6 we removed support for including
> src/localforage.js and require people to include the fully built version
> in dist. I've never officially supported including the source files but I
> think it's always worked until now and a lot of people have been doing
> it--but it causes us headaches [image: :smile:]
>
> I should have marked this as a version bump to 1.3 and didn't. I'll push
> 1.2.7 soon and it should continue working. Then I'll push 1.3, which won't
> work as you've been using but should work with the dist/ version.
>
> Sorry about all this. Should be fixed soon.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/mozilla/localForage/issues/413#issuecomment-127209459>
> .
>
-- 
*Thodoris Greasidis*
[...]

PR #415, opened 2015-08-06
chore: add module wide babel & browserify configs to fix 3rd party builds

### LINKED ISSUE #412: It broke for browserify
Browserify does not check the existence of the "browser-field" in the npm package.json. Even with babelify, it does not work.

thgreasi:
Properly resolves #412 & #413 for the ES6 codebase.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,0 +1,4 @@
+{
+  "loose": "all",
+  "plugins": ["system-import-transformer"]
+}
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -34,12 +34,9 @@ module.exports = exports = function(grunt) {
     grunt.initConfig({
         babel: {
             options: {
-                loose: 'all',
                 modules: 'umd',
                 moduleIds: true,
-                // sourceMap: true,
-                getModuleId: babelModuleIdProvider,
-                plugins: ['system-import-transformer']
+                getModuleId: babelModuleIdProvider
             },
             dist: {
                 files: {
@@ -55,10 +52,8 @@ module.exports = exports = function(grunt) {
             client: {
                 options: {
                     transform: [['babelify', {
-                        loose: 'all',
                         moduleIds: true,
-                        getModuleId: babelModuleIdProvider,
-                        plugins: ['system-import-transformer']
+                        getModuleId: babelModuleIdProvider
                     }]]
                 },
                 src: [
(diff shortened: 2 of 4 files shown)


thgreasi:
The whole point of this PR it to fix localforage for browserify users, before changing the current build system so that we point all users to the `dist` files.

It still points browserify to the `src` files (like before the babel&es6 changes) but discloses the transformations and configurations required to build localforage sources with `browserify` (specifically `babelify` and `system-import-transformer`).

In my tests [open-flight-planner](https://github.com/Turbo87/open-flight-planner) seems to work, requiring only `system-import-transformer`.
@rejas, @Turbo87, @bhamblok can you cross check that your apps work with [this branch](https://github.com/thgreasi/localForage/tree/3rdparty-babelify-build-fix)?

rejas:
thx, will take a look at it tomorrow morning

rejas:
Not sure, maybe I am doing somehting wrong, but I get when using your branch:

[11:03:17] Browserify Error { [ReferenceError: Unknown plugin "system-import-transformer" while parsing file: /Users/mveeck/Projekte/.../node_modules/ampersand-sync-localforage/node_modules/localforage/src/localforage.js]

thgreasi:
**TL;DR** You need to install [system-import-transformer](https://github.com/thgreasi/babel-plugin-system-import-transformer).

This is absolutely expected, since when using browserify to build a `node_module` into your dist, you need to have all the required transformers and plugins installed.
Browserify would also complain in case you were missing babelily itself, since it is [a dependency specified for the browserify build in package.json](https://github.com/thgreasi/localForage/blob/8b34f2904f90ede9267244e685bf39d5d7acfead/package.json#L52-L54).

As I said earlier, this is just a quick fix to allow browserify users to use the code in the master branch again. This is obviously not an ideal solution but just a temporary one. The next step is to point them to a browserify-friendly dist file (unfortunately this was never the case before).

rejas:
Installed it, and it worked fine. :thumbsup:  Thanks for the work, hope it gets merged.

thgreasi:
Just pushed 352d258a44c7996b659a90d3bab672c5b0fe18b4 that modifies the browserify build test, to use the module sources (and build options) as defined in package.json.
Hopefully this will prevent us from breaking localforage for browserify users again.

tofumatt:
Sweeeeet!

This means the test is using the same build that anyone else using browserify would use now, right? :+1:

thgreasi:
Yea, more or less.
Most of the browserify build options have moved to package.json and a babel configuration file. Even in the test case we require the main repo directory, making browserify read package.json for the build options and file paths.
Next thing is to do the same for webpack testing.

Issue #407, opened 2015-07-18
localStorage driver iterate function fetches items which are not owned by localforage db


rasikas:
I found this issue when localforage is used in an environment where browser is IE and when localforage is using localStorage driver.

I think issue is localforage assumes entire localstorage is owned by itself which is not the case in certain applications, some other javascripts and apps can use same localstorage to store their values. So localforage must not assume and work as its own entire localstorage and its data. 

Check the following iterate method

https://github.com/mozilla/localForage/blob/master/src/drivers/localstorage.js#L102
~~~js
    function iterate(iterator, callback) {
        var self = this;

        var promise = self.ready().then(function() {
            var keyPrefix = self._dbInfo.keyPrefix;
            var keyPrefixLength = keyPrefix.length;
            var length = localStorage.length;

            for (var i = 0; i < length; i++) {
                var key = localStorage.key(i);
                var value = localStorage.getItem(key);

                // If a result was found, parse it from the serialized
                // string into a JS object. If result isn't truthy, the
                // key is likely undefined and we'll pass it straight
                // to the iterator.
                if (value) {
                    value = serializer.deserialize(value);
                }

                value = iterator(value, key.substring(keyPrefixLength), i + 1);

                if (value !== void(0)) {
                    return value;
                }
            }
        });

        executeCallback(promise, callback);
        return promise;
    }
~~~
After `var key = localStorage.key(i);` it should check whether key is related to its data before fetching the value. This check has been done in other methods. It should not deserialize the item if it is not stored by localforage (simply someone else' data) which could throw exceptions and unpredictable behaviours, Moreover caller of this library expect to get data what it stored using localforage, not all the data stored in browser localstorage.

fatfisz:
This surprised me too, shouldn't `iterate` and `keys` be consistent in which keys do they ignore?

tofumatt:
Yes, this looks like a bug, I thought it had already been reported but I can't find it (at least on my phone).

Thanks for finding it!

af7:
Hello, this is a bug that causes errors in my case, because I use *two* dbs inside localStorage.

This is why I added this code...

````
if (!(key.indexOf(keyPrefix) === 0)) continue;
````

...between these lines here:

````
var key = localStorage.key(i);
var value = localStorage.getItem(key);
````
https://github.com/mozilla/localForage/blob/master/src/drivers/localstorage.js#L112-L113

Could you *please* add these to MASTER so that localForage behaves as expected?

[...]

PR #418, opened 2015-08-13
Fix localStorage iterate method

af7:
Related to issue #407

### DIFF
--- a/src/drivers/localstorage.js
+++ b/src/drivers/localstorage.js
@@ -110,6 +110,9 @@
 
             for (var i = 0; i < length; i++) {
                 var key = localStorage.key(i);
+                if (key.indexOf(keyPrefix) !== 0) { 
+                    continue;
+                }
                 var value = localStorage.getItem(key);
 
                 // If a result was found, parse it from the serialized


tofumatt:
Thanks! (Was it really that simple, haha?)

Would you be able to write a test for this?

thgreasi:
I think we already have a similar test for the keys() method. We could
probably use the same for iterate().

On Thu, Aug 13, 2015, 18:15 Matthew Riley MacPherson <
notifications@github.com> wrote:

> Thanks! (Was it really that simple, haha?)
>
> Would you be able to write a test for this?
>
> —
> <https://github.com/mozilla/localForage/pull/418#issuecomment-130725914>.
>
-- 
*Thodoris Greasidis*
Computer, Networking &
Communications Engineer

thgreasi:
[This is the test I was talking about](https://github.com/mozilla/localForage/blob/3257d192e47d553f789fe62007f3e6d46aa4b04b/test/test.api.js#L529-L541).

tofumatt:
Oh, nice. Okay, I'm just coming home from the gym. I can do that test in a few hours tops.



- tofumatt

PR #419, opened 2015-08-13
Fix iterate in localstorage issue 407

### LINKED ISSUE #407: localStorage driver iterate function fetches items which are not owned by localforage db

tofumatt:
Adds a test and fixes a minor whitespace fix from #418.

Fixes #407.


af7:
Hello Matthew,

I am not sure if it is totally needed, but you have this final code on the other tests while you do not have it on this test:

`localStorage.clear();`

FYI, this is what I did for a test (I will not do a PR now obviously):
https://github.com/af7/localForage/commit/229d30709f02d7b56faf5a0c2087c959c0c082c3

Thanks.

af7:
Also, in your website docs, you call the argument of the `successCallback` as `result` but you use it as `error` on the test.

This is what you have on the Docs...

````
// Exit the iteration early:
    if (iterationNumber < 3) {
        console.log([key, value]);
    } else {
        return [key, value];
    }
}, function(result) {
    console.log('Iteration has completed, last iterated pair:');
    console.log(result);
});
````

...and this is what you use on the Test:

````
}, function(err) {
     if (!err) {
````

tofumatt:
It isn't needed in that test (it might not even be needed in the other ones actually, as localForage should always stick to its own namespace and not need to clean out ALL of localStorage).

Our tests are pretty similar, thanks for sharing. As you see: it's more about checking to make sure "out-of-bounds" items _don't_ get returned than checking for what is returned.

Thanks!

tofumatt:
You're right, though that's a separate issue. I'll check into that...

tofumatt:
fwiw, looks like that’s an error in the docs. I will fix it and push that now. Thanks for catching it!

- tofumatt

af7:
Concerning the `localStorage.clear();` question, I thought that this is done so that the **next** test starts with a clean localStorage environment (which is needed for proper testing). This is why I though it is kind of necessary.

It seems that tests that need this *clean* environment could start with `localStorage.clear();`, so that you do not have to do this at the end of each test. In any case, you know much better what needs to be done (or not).

Thanks again!

tofumatt:
Turns out localStorage.clear() is run in the `beforeEach` block. That clear() in the actual method might even be legacy.


Thanks for reporting. :-)


- tofumatt

PR #422, opened 2015-08-14
Docs: Fix the iterate callback arguments

af7:


af7:
Also, in the edited example, what if the db has just 2 keys?
The `result` on the `SuccessCallback` will be `undefined` I assume? So the message "Iteration has completed, last iterated pair:" does not make much sense in this case?

tofumatt:
Oops, can't believe I missed this, thanks!

That's true, but it's just supposed to be a quick example... you could make it more accurate, but it would make it longer. For completeness, you'd want:

```javascript

// Exit the iteration early:
    if (iterationNumber < 3) {
        console.log([key, value]);
    } else {
        return [key, value];
    }
}, function(err, result) {
    if (!err) {
        if (result) {
            console.log('Iteration has completed, last iterated pair:');
            console.log(result);
        } else {
            console.log('Iteration has completed.');
        }
    }
});

PR #428, opened 2015-08-17
Fix the 'iterationNumber' inconsistency of localStorage iterate method

af7:
I would expect that the iterationNumber starts from 1 and increments only by 1 for each item found on localStorage for the specific localForage instance.

For example, in my use case I have 2 different instances of localForage on localStorage, plus other non-localForage related items on localStorage. 

Unfortunately, the iterationNumber is related to the order of the item on the entire localStorage, so there are gaps on its value when iterating on a specific instance of localForage. And this is not useful at all.

Note: Feel free to change the test to your liking, that was a quick shot.

### DIFF
--- a/src/drivers/localstorage.js
+++ b/src/drivers/localstorage.js
@@ -107,6 +107,7 @@
             var keyPrefix = self._dbInfo.keyPrefix;
             var keyPrefixLength = keyPrefix.length;
             var length = localStorage.length;
+            var iterationNumber = 1;
 
             for (var i = 0; i < length; i++) {
                 var key = localStorage.key(i);
@@ -123,7 +124,7 @@
                     value = serializer.deserialize(value);
                 }
 
-                value = iterator(value, key.substring(keyPrefixLength), i + 1);
+                value = iterator(value, key.substring(keyPrefixLength), iterationNumber++);
 
                 if (value !== void(0)) {
                     return value;
--- a/test/test.api.js
+++ b/test/test.api.js
@@ -261,6 +261,9 @@ DRIVERS.forEach(function(driverName) {
         });
 
         it('should iterate() through only its own keys/values', function(done) {
+
+            var iterationNumberConcat = '';
+
             localStorage.setItem('local', 'forage');
 
             localforage.setItem('office', 'Initech').then(function() {
@@ -269,16 +272,21 @@ DRIVERS.forEach(function(driverName) {
                 // Loop through all key/value pairs; local: 'forage' set
                 // manually should not be returned.
                 var numberOfItems = 0;
-                localforage.iterate(function(value, key) {
+                localforage.iterate(function(value, key, iterationNumber) {
                     // Returning defined value will break the cycle.
                     expect(key).to.not.be('local');
                     expect(value).to.not.be('forage');
                     numberOfItems++;
+                    iterationNumberConcat += iterationNumber;
                 }, function(err) {
                     if (!err) {
                         // Only two items were set, so that's all we should
                         // have!
                         expect(numberOfItems).to.be(2);
+
+                        // Only one item was set on localForage storage,
+                        // so we should get '1' and not '12'
+                        expect(iterationNumberConcat).to.be('1');
 
                         done();
                     }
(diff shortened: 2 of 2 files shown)


af7:
Also, shouldn't this 'iterate' test be in the section where the driver is 'localStorage' only?

tofumatt:
It doesn't hurt to be in this section, so I'm fine with it there. I'm surprised we _didn't_ already have a test for it, but maybe I'm confused as to what the issue was in the first place. Could you provide some example code and how it behaves differently with your patch? I think I get what it does but I'm not quite sure!

tofumatt:
Are the tests all passing on your machine? PhantomJS is timing out on Travis, but I can't tell if that's a Travis bug or not :smile:

af7:
The example code is the own you have in your documentation for 'iterate', just add some localStorage keys that do not have the prefix of localForage, preferably with characters that will make them go both above and below the order of localStorage.

Then, perform the iterate() and you will see that the iterationNumber increments also for these items that are ignored by the loop (with the 'continue').

af7:
For tests, I can't really tell, you may disregard my test. I am on a windows machine and had problems setting up the environment, I always get messages like this on most tests:

Warning: PhantomJS timed out, possibly due to a missing Mocha
 run() call. Use --force to continue.

af7:
I believe that this fix is definitely needed along with the other one (that made it to the master). This way the localStorage driver is consistent with the other 2 drivers (no gaps in iterationNumber).

I hope you can add this and release the 1.3 version to NPM, as these fixes in localStorage driver makes it possible for us to use localForage. Thanks!

tofumatt:
Yeah, I really appreciate you battle-testing the localStorage components more—not many people seem to use it there but I agree, these are important!




I’m a bit swamped now and we are trying to make sure our browserify/webpack users get a good release in 1.3 so we might be delayed a bit, but know that I’m working to get this stuff out ASAP. It should be this week, if not I’ll back port these fixes into the pre-es6 branch 1.2.7 is based on :-)


- tofumatt

af7:
Hell Matthew, can you please add the code to master, the tests were failing due to "timeout" and don't know why (the code is very important to be added in v1.3). How can we re-run the tests? (I see no such option in the interface).

tofumatt:
If the tests are timing out, it could have been a Travis issue (I restarted them), but I will run them locally as well. Our tests should pass without `--force`, so if it's timing out it's due to a test problem.

I can give this more time soon but it's a bit of a hectic week. Sorry I couldn't get to it before.

Are the tests running locally? We do have support for Windows, so the tests should run on your machine if you have node installed. If you can't run the tests I would encourage you to file bugs about that so we can fix it, as that's quite important!

tofumatt:
I just ran the tests locally and was treated to an error. I'll give it a quick look now but I only have about thirty minutes tonight :frowning:

tofumatt:
Okay, I just looked through this branch, but you've misunderstood the test I believe. There are three items set in localStorage, two of which are set by localForage:

```js
        it('should iterate() through only its own keys/values', function(done) {
            localStorage.setItem('local', 'forage');
            localforage.setItem('office', 'Initech').then(function() {
                // ###### Note that we set a second item here ######
                return localforage.setItem('name', 'Bob');
            }).then(function() {
                // Loop through all key/value pairs; local: 'forage' set
                // manually should not be returned.
                var numberOfItems = 0;
                    // Returning defined value will break the cycle.
                    expect(key).to.not.be('local');
                    expect(value).to.not.be('forage');
                    numberOfItems++;
                }, function(err) {
                    if (!err) {
                        // While there are 3 items in localStorage,
                        // only two items were set using localForage.
                        expect(numberOfItems).to.be(2);

                        done();
                    }
                });
            });
        });
```

The docs in the comments were bad, however, and previously referred to only one thing being set by localForage. I've fixed that up in master and closed this issue here.

af7:
Yes, you are right, i will create a new test and submit a PR again. Note that without this change the solution for localStorage is not working! So, i am afraid that a 1.2.9 should be released on NPM after I send you  the new PR and tests do not fail.

P.S. I have it ready, but I get a "Failed to connect to github.com port 443: Timed out" when trying to push to github... I hope this is a temporary problem, I will try again later.

af7:
Here is the new PR: https://github.com/mozilla/localForage/pull/434

Note that without this change the solution for localStorage in 1.2.8 is not working... So, I am afraid that a 1.2.9 should be released on NPM with the change of PR434 included.

Thanks!

PR #429, opened 2015-08-19
test(iterate): add 'done()' to spec

af7:


tofumatt:
Surprised the test ran before without this; looks right to me. Thanks for catching it.

PR #434, opened 2015-08-25
Fix the iterationNumber logic on localStorage

af7:


af7:
Note that without this change the solution for localStorage in 1.2.8 is not working... So, I am afraid that a 1.2.9 should be released on NPM with the change of PR434 included.

Thanks!

tofumatt:
Okay, all sorted now! I've merged this into 1.2 and release 1.2.9 with the fix.

af7:
Thanks! Now time to pack for your flight, enjoy!

PR #435, opened 2015-08-26
fix(localStorageDriver.iterate): fix the iterationNumber logic

af7:
I added this *necessary* addition to the localStorage's `iterate` fix to the `master` too (it was added to the `pre-ES6` branch a day ago), so that it is also included to the 1.3 release.

I hope that you can also merge the other PR to the `pre-ES6` and upload a 1.2.9 version on NPM that includes this fix.

Thanks!

### DIFF
--- a/src/drivers/localstorage.js
+++ b/src/drivers/localstorage.js
@@ -107,6 +107,7 @@
             var keyPrefix = self._dbInfo.keyPrefix;
             var keyPrefixLength = keyPrefix.length;
             var length = localStorage.length;
+            var iterationNumber = 1;
 
             for (var i = 0; i < length; i++) {
                 var key = localStorage.key(i);
@@ -123,7 +124,7 @@
                     value = serializer.deserialize(value);
                 }
 
-                value = iterator(value, key.substring(keyPrefixLength), i + 1);
+                value = iterator(value, key.substring(keyPrefixLength), iterationNumber++);
 
                 if (value !== void(0)) {
                     return value;
--- a/test/test.api.js
+++ b/test/test.api.js
@@ -262,6 +262,7 @@ DRIVERS.forEach(function(driverName) {
         });
 
         it('should iterate() through only its own keys/values', function(done) {
+
             localStorage.setItem('local', 'forage');
             localforage.setItem('office', 'Initech').then(function() {
                 return localforage.setItem('name', 'Bob');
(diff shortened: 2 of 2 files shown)


src/drivers/localstorage.js
```diff
@@ -123,7 +124,7 @@
                     value = serializer.deserialize(value);
                 }
 
-                value = iterator(value, key.substring(keyPrefixLength), i + 1);
+                value = iterator(value, key.substring(keyPrefixLength), iterationNumber++);
```
tofumatt: Why this change? You can see above we use `i` as a zero-index through the keys, so adding one to it then returning does the same thing as returning your 1-indexed variable.

tofumatt:
Sorry, but what exactly is the use case that is failing for you using localStorage? Could you provide a code example? I added your tests without changing the localStorage driver in `src/drivers/localstorage.js` and the tests all passed.

Do you think the `iterationNumber` isn't accurate? I don't undestand the need for these changes.

src/drivers/localstorage.js
thgreasi: Probably to keep out of count the items fall in the `continue` case (L115).

tofumatt: Ahhhh, okay. You're right, I was being dense.

Oddly the test doesn't actually test for it, so I'll rework it and merge it in afterward. Thanks :smiley:

thgreasi: I think that the test do not suffice only because of the internal implementation of locastorage (probably browser specific).
It seems that even though `localStorage.setItem('local', 'forage');` is executed first, `iterate` returns it last !?
In my tests in chrome (on this exact page if that matters) `localStorage.key(0)` returns the last item inserted, like a stack.
Adding an  extra item in localStorage right after `localforage.setItem('name', 'Bob');` resolves should be enough.

tofumatt: Weird! Interesting. Okay, I'll fix that up and try to do it now. I've gotta pack for holiday though, so if I don't do it now I'll get to it tonight :smile:

af7: Thank you for your efforts on this matter.

Indeed, the problem was that we should skip the non-localForage items from
the iterationNumber count.

Also, I have noticed the same as Thodoris about the order of items that the
`iterate` brings, it seems to be the opposite order (it does not affect the
iterationNumber count though), so the test is OK.

tofumatt:
Okay, all sorted now! I've merged this into 1.2 and release 1.2.9 with the fix.

I'll get this patch into master soon but I need to catch a flight first :wink: :airplane:

tofumatt:
Sorry, I ended up taking a real vacation away from computers and this week has been busy. Finally merged into master, thanks so much for finding the bugs and these fixes/tests @af7, really appreciated! :+1:

af7:
You're welcome! :-)

PR #452, opened 2015-09-25
Fix #342 : multiple stores on the same database

### LINKED ISSUE #342: localforage.createInstance: cannot create several instances with the same name (=db name)

sebweaver:
Implementation summary for IndexedDB:
- Keep track on all running forages per database.
- Wait for them to be ready before processing any new connection to the same
database.
- Probe the current database to check if an upgrade is required (greater version
or new store).
- If so, upgrade the database by reopening it.
- Share the final connection (original or upgraded) amongts all running forages.

Implementation summary for localStorage:
- Prefix the key with the store name (after database), unless it's the default
name.

### DIFF
--- a/src/drivers/indexeddb.js
+++ b/src/drivers/indexeddb.js
@@ -206,11 +206,18 @@
         dbContext.forages.push(this);
 
         // Create an array of readiness of the related localForages.
-        var ready = [];
+        var readyPromises = [];
+
+        function ignoreErrors() {
+            // Don't handle errors here,
+            // just makes sure related localForages aren't pending.
+            return Promise.resolve();
+        }
+
         for (var j = 0; j < dbContext.forages.length; j++) {
             var forage = dbContext.forages[j];
             if (forage !== this) { // Don't wait for itself...
-                ready.push(forage.ready());
+                readyPromises.push(forage.ready().catch(ignoreErrors));
             }
         }
 
@@ -218,8 +225,8 @@
         var forages = dbContext.forages.slice(0);
 
         // Initialize the connection process only when
-        // all the related localForages are ready.
-        return Promise.all(ready).then(function() {
+        // all the related localForages aren't pending.
+        return Promise.all(readyPromises).then(function() {
             dbInfo.db = dbContext.db;
             // Get the connection or open a new one without upgrade.
             return _getOriginalConnection(dbInfo);
--- a/src/drivers/indexeddb.js
+++ b/src/drivers/indexeddb.js
@@ -181,25 +181,162 @@
             }
         }
 
+        // Initialize a singleton container for all running localForages.
+        if (!globalObject.container) {
+            globalObject.container = {};
+        }
+
+        // Get the current context of the database;
+        var context = globalObject.container[dbInfo.name];
+
+        // ...or create a new context.
+        if (!context) {
+            context = {
+                // Running localForages sharing a database.
+                forages: [],
+                // Shared database.
+                db: null
+            };
+            // Register the new context in the global container.
+            globalObject.container[dbInfo.name] = context;
+        }
+
+        // Register itself as a running localForage in the current context.
+        context.forages.push(this);
+
+        // Create an array of readiness of the related localForages.
+        var ready = [];
+        for (var j = 0; j < context.forages.length; j++) {
+            var forage = context.forages[j];
+            if (forage !== this) { // Don't wait for itself...
+                ready.push(forage.ready());
+            }
+        }
+
+        // Take a snapshot of the related localForages.
+        var forages = context.forages.slice(0);
+
+        // Initialize the connection process only when
+        // all the related localForages are ready.
+        return Promise.all(ready).then(function() {
+            dbInfo.db = context.db;
+            // Get the connection or open a new one without upgrade.
+            return _getOriginalConnection(dbInfo);
+        }).then(function(db) {
+            dbInfo.db = db;
+            if (_isUpgradeNeeded(dbInfo, self._defaultConfig.version)) {
+                // Reopen the database for upgrading.
+                return _getUpgradedConnection(dbInfo);
+            }
+            return db;
+        }).then(function(db) {
+            dbInfo.db = context.db = db;
+            self._dbInfo = dbInfo;
+            // Share the final connection amongst related localForages.
+            for (var k in forages) {
+                var forage = forages[k];
+                if (forage !== self) { // Self is already up-to-date.
+                    forage._dbInfo.db = dbInfo.db;
+                    forage._dbInfo.version = dbInfo.db;
+                }
+            }
+        });
+    }
+
+    function _getOriginalConnection(dbInfo) {
+        return _getConnection(dbInfo, false);
+    }
+
+    function _getUpgradedConnection(dbInfo) {
+        return _getConnection(dbInfo, true);
+    }
+
+    function _getConnection(dbInfo, upgradeNeeded) {
         return new Promise(function(resolve, reject) {
-            var openreq = indexedDB.open(dbInfo.name, dbInfo.version);
+            if (dbInfo.db) {
+                if (upgradeNeeded) {
+                    dbInfo.db.close();
+                } else {
+                    return resolve(dbInfo.db);
+                }
+            }
+
+            var dbArgs = [dbInfo.name];
+
+            if (upgradeNeeded) {
+                dbArgs.push(dbInfo.version);
+            }
+
+            var openreq = indexedDB.open.apply(indexedDB, dbArgs);
+
+            if (upgradeNeeded) {
+                openreq.onupgradeneeded = function(e) {
+                    var db = openreq.result;
+                    try {
+                        db.createObjectStore(dbInfo.storeName);
+                        if (e.oldVersion <= 1) {
+                            // Added when support for blob shims was added
+                            db.createObjectStore(DETECT_BLOB_SUPPORT_STORE);
+                        }
+                    } catch (ex) {
+                        if (ex.name === 'ConstraintError') {
+                            globalObject.console.warn('The database "' + dbInfo.name + '"' +
+                                ' has been upgraded from version ' + e.oldVersion +
+                                ' to version ' + e.newVersion +
+                                ', but the storage "' + dbInfo.storeName + '" already exists.');
+                        } else {
+                            throw ex;
+                        }
+                    }
+                };
+            }
+
             openreq.onerror = function() {
                 reject(openreq.error);
             };
-            openreq.onupgradeneeded = function(e) {
-                // First time setup: create an empty object store
-                openreq.result.createObjectStore(dbInfo.storeName);
-                if (e.oldVersion <= 1) {
-                    // added when support for blob shims was added
-                    openreq.result.createObjectStore(DETECT_BLOB_SUPPORT_STORE);
-                }
-            };
+
             openreq.onsuccess = function() {
-                dbInfo.db = openreq.result;
-                self._dbInfo = dbInfo;
-                resolve();
+                resolve(openreq.result);
             };
         });
+    }
+
+    function _isUpgradeNeeded(dbInfo, defaultVersion) {
+        if (!dbInfo.db) {
+            return true;
+        }
+
+        var isNewStore = !dbInfo.db.objectStoreNames.contains(dbInfo.storeName);
+        var isDowngrade = dbInfo.version < dbInfo.db.version;
+        var isUpgrade = dbInfo.version > dbInfo.db.version;
+
+        if (isDowngrade) {
+            // If the version is not the default one
+            // then warn for impossible downgrade.
+            if (dbInfo.version !== defaultVersion) {
+                globalObject.console.warn('The database "' + dbInfo.name + '"' +
+                    ' can\'t be downgraded from version ' + dbInfo.db.version +
+                    ' to version ' + dbInfo.version + '.');
+            }
+            // Align the versions to prevent errors.
+            dbInfo.version = dbInfo.db.version;
+        }
+
+        if (isUpgrade || isNewStore) {
+            // If the store is new then increment the version (if needed).
+            // This will trigger an "upgradeneeded" event which is required
+            // for creating a store.
+            if (isNewStore) {
+                var incVersion = dbInfo.db.version + 1;
+                if (incVersion > dbInfo.version) {
+                    dbInfo.version = incVersion;
+                }
+            }
+
+            return true;
+        }
+
+        return false;
     }
 
     function getItem(key, callback) {
--- a/src/drivers/indexeddb.js
+++ b/src/drivers/indexeddb.js
@@ -16,6 +16,7 @@
 
     var DETECT_BLOB_SUPPORT_STORE = 'local-forage-detect-blob-support';
     var supportsBlobs;
+    var dbContexts;
 
     // Abstracts constructing a Blob object, so it also works in older
     // browsers that don't support the native Blob constructor. (i.e.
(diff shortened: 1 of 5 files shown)


thgreasi:
:+1:

src/drivers/indexeddb.js
```diff
@@ -181,27 +181,164 @@
             }
         }
 
+        // Initialize a singleton container for all running localForages.
+        if (!globalObject.container) {
+            globalObject.container = {};
```
thgreasi: Is there a reason that we attach the `container` on the global object?? It feels too intrusive to me to just add an extra variable to the window object.

This could probably be just a local variable [around line 18](https://github.com/sebweaver/localForage/blob/issue-342-multiple-stores-on-same-db/src/drivers/indexeddb.js#L18), inside the closure of the IndexedDB driver.

As an alternative, it could be attached on the LocalForage class as a static property, named like `_dbContexts`. On the other hand I don't see a reason that this couldn't just be completely private to the IndexedDB driver.

src/drivers/localstorage.js
```diff
@@ -39,6 +39,10 @@
 
         dbInfo.keyPrefix = dbInfo.name + '/';
 
+        if (dbInfo.storeName !== self._defaultConfig.storeName) {
+            dbInfo.keyPrefix += dbInfo.storeName + '/';
+        }
```
thgreasi: Simply said, :+1: on this. I can't remember which issue it closes!?

test/test.config.js
```diff
@@ -197,7 +197,7 @@ describe('Config API', function() {
                     });
                 });
             } else if (localforage.driver() === localforage.LOCALSTORAGE) {
-                var dbValue = JSON.parse(localStorage['My Cool App/some key']);
+                var dbValue = JSON.parse(localStorage['My Cool App/storeFront/some key']);
```
thgreasi: I would prefer not to use the 'storeFront' string in here, since it is related to several trademarked products.

src/drivers/indexeddb.js
sebweaver: My bad, you're totally right about `globalObject`! It's the first thing I did on this PR and I forgot to change it. I will attach this container locally in the IndexedDB driver, name it `dbContexts`, as you suggest, and update the PR.

src/drivers/localstorage.js
sebweaver: No issue in particular AFAIK, but necessary to fulfill the new test case of this PR.

In fact, it's surprising the support of `storeName` hasn't been implemented before. At least in the same time than the two other drivers. In any case, I didn't see anything about that in the localStorage driver.

test/test.config.js
sebweaver: Not my choice of name. :-)

I've just adapted the test to make it pass with the implementation of the `storeName` namespace of this PR. The original name was here:

https://github.com/sebweaver/localForage/blob/issue-342-multiple-stores-on-same-db/test/test.config.js#L78

thgreasi: Ok then!

src/drivers/localstorage.js
thgreasi: I remember someone had reported this somewhere....
Anyway, did you also add a test case for this?

sebweaver: Test cases already exist here:

https://github.com/sebweaver/localForage/blob/issue-342-multiple-stores-on-same-db/test/test.api.js#L772

And there:

https://github.com/sebweaver/localForage/blob/issue-342-multiple-stores-on-same-db/test/test.api.js#L799

I just enriched them and they failed for localStorage too (not only IndexedDB). That's why I added the above implementation. Far simpler than for IndexedDB by the way... :-)

thgreasi: Great :+1:

src/drivers/indexeddb.js
```diff
@@ -181,27 +182,164 @@
...
+            self._dbInfo = dbInfo;
+            // Share the final connection amongst related localForages.
+            for (var k in forages) {
+                var forage = forages[k];
+                if (forage !== self) { // Self is already up-to-date.
+                    forage._dbInfo.db = dbInfo.db;
+                    forage._dbInfo.version = dbInfo.db;
+                }
+            }
+        });
```
thgreasi: Should we also add a `catch()` here and `reject` the promise returned by `_initStorage`?

sebweaver: No, there's no nested promise here, but a chain of existing promises which have been already resolved of rejected individually. If a rejection occurs in a chain of promises, it bubbles up and the whole chain is rejected.

thgreasi: Hmmmm
Lets say I create instance1 giving a very large amount of DB size and as a result the ready() promise of instance1 rejects at some point.
Now lets say that my code handles that error and creates instance2 with the same db parameters as instance1 but with a smaller & valid size for the db size option. 
Will that work? I think not, since we wait until all `ready()` promises of the related instances are **resolved** (even though your comment says "Initialize the connection process only when all the related localForages are ready").
I think that we just need to know that their promises are no longer pending and as a result their `_iniStorage` call has completely finished.

TL;DR: I think that instead of `ready.push(forage.ready());` in [Line 213](https://github.com/mozilla/localForage/pull/452/files#diff-29be772a36519980595b932d60929feeR213) we should use `ready.push(forage.ready().catch())` (if I'm not mistaken or understood anything wrong);

test/test.config.js
tofumatt: We can certainly change it. I wrote that so long ago I can't remember why, and it's worth changing ;-)

I'll do that outside this PR (or when we merge it in): #453 is filed for it :smile:

tofumatt:
Wow, this is _awesome_! :+1:

src/drivers/indexeddb.js
```diff
@@ -181,27 +182,164 @@
...
+            // Register the new context in the global container.
+            dbContexts[dbInfo.name] = dbContext;
+        }
+
+        // Register itself as a running localForage in the current context.
+        dbContext.forages.push(this);
+
+        // Create an array of readiness of the related localForages.
+        var ready = [];
+        for (var j = 0; j < dbContext.forages.length; j++) {
```
tofumatt: Any reason you're using `j` over `i`? ^_^

src/drivers/indexeddb.js
tofumatt: If there's no harm to @thgreasi's suggestion I _think_ I agree with him, but I'll admit my brain gets scrambled by Promises sometimes at this level.

If it would be possible to write a test to cover @thgreasi's use-case that would be great. Would it be as simple as:

```js
var databaseInfo = {
    name: 'localForaging',
    storeName: 'helloApp',
    size: 24903680
};

var db = localforage.createInstance(databaseInfo)

db.then(function() {
    // This should fail?
}, function() {
    databaseInfo.size = 5000;
    db = localforage.createInstance(databaseInfo);
});

Kind of a contrived example, but it should do the trick? Unless I misunderstand! ^_^

**Comment by tofumatt** [comment]
Other than the question of using `catch()` as @thgreasi mentioned here: https://github.com/mozilla/localForage/pull/452#discussion_r40494119

this is awesome and gets a :+1: r+ from me.

Once we figure out what needs to me done there we can merge it. I have a few nitpicks about variable names and stuff but we can fix those after merge; this PR is great :smile: 

Thanks @sebweaver!

File: src/drivers/indexeddb.js (hunk 4 above)
Comment (thgreasi) [inline src/drivers/indexeddb.js]: Hmmm.... (again),
Changing `ready.push(forage.ready());` in [line 213](https://github.com/mozilla/localForage/pull/452/files#diff-29be772a36519980595b932d60929feeR213) into `ready.push(forage.ready().catch());` should fix the aforementioned test case...
Also we should probably add a comment there explaining that we only care about the promise not being pending.
Too much effort for such a small change :smile:

Comment (tofumatt) [inline src/drivers/indexeddb.js]: Should we just add in the `catch` and then merge this into master and address this use case later if needed then?

I'd be okay with that... this patch is awesome.

Comment (thgreasi) [inline src/drivers/indexeddb.js]: Now you made me hesitant about the completeness and correctness of my
suggested fix...
This is a way too small test case to leave for an other PR. On the other
hand I feel ok with the current state of this PR.
How about giving it an extra day or two?

On Sun, Sep 27, 2015, 22:16 Matthew Riley MacPherson <
notifications@github.com> wrote:

> In src/drivers/indexeddb.js
> <https://github.com/mozilla/localForage/pull/452#discussion_r40508172>:
>
> > +                // Reopen the database for upgrading.
> > +                return _getUpgradedConnection(dbInfo);
> > +            }
> > +            return db;
> > +        }).then(function(db) {
> > +            dbInfo.db = dbContext.db = db;
> > +            self._dbInfo = dbInfo;
> > +            // Share the final connection amongst related localForages.
> > +            for (var k in forages) {
> > +                var forage = forages[k];
> > +                if (forage !== self) { // Self is already up-to-date.
> > +                    forage._dbInfo.db = dbInfo.db;
> > +                    forage._dbInfo.version = dbInfo.db;
> > +                }
> > +            }
> > +        });
>
> Should we just add in the catch and then merge this into master and
> address this use case later if needed then?
>
> I'd be okay with that... this patch is awesome.
>
> —
> <https://github.com/mozilla/localForage/pull/452/files#r40508172>.
>
-- 
*Thodoris Greasidis*
Computer, Networking &
Communications Engineer

Comment (tofumatt) [inline src/drivers/indexeddb.js]: Cool, I can be patient. I’ll wait for a bit and check back in when it’s figured out :-)

- tofumatt

On 27 September 2015 at 20:26:39, Thodoris Greasidis (notifications@github.com) wrote:

In src/drivers/indexeddb.js:

> +                // Reopen the database for upgrading.
> +                return _getUpgradedConnection(dbInfo);
> +            }
> +            return db;
> +        }).then(function(db) {
> +            dbInfo.db = dbContext.db = db;
> +            self._dbInfo = dbInfo;
> +            // Share the final connection amongst related localForages.
> +            for (var k in forages) {
> +                var forage = forages[k];
> +                if (forage !== self) { // Self is already up-to-date.
> +                    forage._dbInfo.db = dbInfo.db;
> +                    forage._dbInfo.version = dbInfo.db;
> +                }
> +            }
> +        });
Now you made me hesitant about the completeness and correctness of my suggested fix... This is a way too small test case to leave for an other PR. On the other hand I feel ok with the current state of this PR. How about giving it an extra day or two?
On Sun, Sep 27, 2015, 22:16 Matthew Riley MacPherson < notifications@github.com> wrote: In src/drivers/indexeddb.js <https://github.com/mozilla/localForage/pull/452#discussion_r40508172>: > + // Reopen the database for upgrading. > + return _getUpgradedConnection(dbInfo); > + } > + return db; > + }).then(function(db) { > + dbInfo.db = dbContext.db = db; > + self._dbInfo = dbInfo; > + // Share the final connection amongst related localForages. > + for (var k in forages) { > + var forage = forages[k]; > + if (forage !== self) { // Self is already up-to-date. > + forage._dbInfo.db = dbInfo.db; > + forage._dbInfo.version = dbInfo.db; > + } > + } > + }); Should we just add in the catch and then merge this into master and address this use case later if needed then? I'd be okay with that... this patch is awesome. — Reply to this email directly or view it on GitHub <https://github.com/mozilla/localForage/pull/452/files#r40508172>.
-- *Thodoris Greasidis* Computer, Networking & Communications Engineer

Comment (sebweaver) [inline src/drivers/indexeddb.js]: My first idea was indeed to wait for other localForages to be ready, and by "ready" I mean "fulfilled" (sorry if my comment in the code wasn't clear enough about that). This goal of this implementation is to act as a kind of "transaction" actually. Thus, if several instance are created on the same database and one of them fails (like in @thgreasi's example) this implementation prevents, indeed, the next ones to be created. The idea behind that is to avoid any side-effect or inconsistency which might result from letting the process continuing. I mean, if multiples stores are involved in the same database, they are highly probably related somehow. And if an error is not explicitly handled at the initialization time, it has to be handled at a later time anyway.

That being said, I see your point. While my solution forces users to handle errors at initialization time, your solution to wait for localForages to be "not pending" (i.e. fulfilled or rejected) is indeed less restrictive and let them decide where and when to handle errors. So, on second thought, you might be right.

However, the way you suggest to achieve this goal can't work. Just adding an empty `catch()` on a promise doesn't change the status of the latter and returns it as is (i.e fulfilled or rejected). So the promise array will be globally fulfilled or rejected in the same way as before. On the other hand, a solution would be to return a resolved promise in both cases with:

```javascript
ready.push(forage.ready().catch(function() {
    return Promise.resolve();
});
```

Whereas this pattern could be confusing, if not commented, it gets the job done.

Now, the question which has to be decided is: what behavior do we want for users?

*Note: I hope my english is clear enough (here or in the code), it's not my native language...*

src/drivers/indexeddb.js
sebweaver: Because `i` is already used above and jshint complains about that during the tests (with or without `var`). Since I didn't want to change the jshint config, I changed the name. :-)

sebweaver:
> I have a few nitpicks about variable names and stuff

Sure, feel free to change them, or ask me to change them. English is not my native language, I won't be offended and be happy to improve things... ;-)

src/drivers/indexeddb.js
thgreasi: Actually, [according to mdn](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/catch) `after a catch the chain is restored`, if I understood it right.

sebweaver: I admit it's a bit confusing, but "restoring the chain" doesn't mean changing the status of the original promise. And thankfully so, otherwise one can't be able to start different (and concurrent) chains on the same promise without massive side-effects between them. Promises are only, and exclusively, resolved or rejected by the executor given at their initialization.

Here is a [jsbin](https://jsbin.com/wujaki/edit?js,console) which illustrates this. It's based on the MDN example above and completed with our two arrays of promises. As one can see, even though `catch()` restores all chains in which it's been involved, only the second array `pAry2` is actually resolved.

thgreasi: You are right about `p2.catch().then(function(){ console.log('asdf'); })` not working
I never meant that a `p2.catch()` would alter the promise stored in variable `p2` (if not reassigned).

The fact is that `.catch()` will always return a new Promise so `p2 != p2.catch()`.

The part that confused me, was that it will not change the returned promise status, unless a function is passed as parameter (**any function!**).

In other words `Promise.reject('error').catch(function(){})` will always be resolved.
[Here is a jsbin with the above](https://jsbin.com/cofipapozi/edit?js,console)!?!?

*PS: That's a great discussion for a single PR!*

sebweaver: > I never meant that a p2.catch() would alter the promise stored in variable p2 (if not reassigned).

> The fact is that .catch() will always return a new Promise so p2 != p2.catch().

Fine! In fact, I just based my reasoning on your original suggestion using an empty catch vs. returning a new resolved promise. Sorry if inferred things that should not be. :blush: 

For the record, I've updated this [jsbin](https://jsbin.com/wujaki/edit?js,console) with the new case.

Now we are sure that en empty catch won't do the job unless any function is passed, the main question still remains: should the initialization process wait for others related instances to be fulfilled (current status of this PR) or just not pending?

As I said before, on second thought the second solution appears to be more flexible to users. If accepted, I can update the PR very soon.

thgreasi: I think that it should be @tofumatt 's decision. My call would be to go for the second case, where we check for `not pending`.
If that's how you will proceed, consider using `.catch(() => Promise.resolve())` for simplicity and consistency, as [found in other places in localForage source](https://github.com/mozilla/localForage/blob/086a344212124075cfba114caac1d842df78e5a0/src/localforage.js#L348).

tofumatt: (Sorry for my late reply). I'm in agreement here; the second solution is the more flexible option for developers and I think we should go with it.

Let me know when the PR is updated and I'll have one last look. :+1:

sebweaver: PR updated!

PR #457, opened 2015-10-02
feat: add getSerializer and expose it on driver's dbInfo when used

### LINKED ISSUE #398: Refactor: move serializer variable into dbInfo
Such a change would allow us to make custom drivers DRYer.
Simple drivers could greatly reduce their code and file size by overriding only the different methods and reusing the rest ones. The problem currently is that the serializer is referenced by the closures of the  driver methods. Attaching it to dbInfo would be great.

For example, sessionStorage and CordovaSQLitePlugin drivers would just need to override the _initStorage methods and reuse localStorageWrapper and websqldriver respectively.

Making localforage drivers architecture more object oriented (maybe es6 class based) would be an other step forward.

No PR yet, just discussing. Waiting for other major PRs to merge.

### LINKED ISSUE #446: Expose the serializer on the localForage object/namespace
This is to gather the discussion that appeared in #412 & #432 (@tofumatt, @af7 )
My two suggestions:
* `localforage.serializer` returning the actual serializer object
* `localforage.getSerializer()` returning a promise

## `localforage.serializer` returning the actual serializer object
Pros:
* use es6 static imports

Cons:
* load the serializer even in cases that it's not needed (eg: IndexedDB with unbundled RequireJS, do we even support it any more?)

This will make even more sense if we move [the IndexedDB driver's blob serialization code](https://github.com/mozilla/localForage/blob/9d2b28b5f508f2dd38e4a133ddd5ca9baaea7209/src/drivers/indexeddb.js#L17-L168) into the main serializer.

Sample implementation:
```js
// on the top of localforage.js, ouside the IIFE
import serializer from 'utils/serializer'

// near the end of localforage.js inside the IIFE
localForage.serializer = serializer;
export default localForage;
```

## `localforage.getSerializer()` returning a promise
Pros:
* We don't import anything if not really required/requested
* That's the way the drivers are currently implemented and we could refactor them to use the exposed method, instead of doing a `System.import` in their own

Cons:
* The developer will have to wait for a promise to resolve

```js
// add a method to the LocalForage class
getSerializer() {
  return System.import('./utils/serializer');
}
```

thgreasi:
Resolves #398
Resolves #446

### DIFF
--- a/src/drivers/localstorage.js
+++ b/src/drivers/localstorage.js
@@ -6,7 +6,6 @@
     'use strict';
 
     var globalObject = this;
-    var serializer = null;
     var localStorage = null;
 
     // If the app is running inside a Google Chrome packaged webapp, or some
@@ -42,7 +41,7 @@
         self._dbInfo = dbInfo;
 
         return System.import('./../utils/serializer').then(function(lib) {
-            serializer = lib;
+            dbInfo.serializer = lib;
             return Promise.resolve();
         });
     }
@@ -89,7 +88,7 @@
             // is likely undefined and we'll pass it straight to the
             // callback.
             if (result) {
-                result = serializer.deserialize(result);
+                result = dbInfo.serializer.deserialize(result);
             }
 
             return result;
@@ -104,7 +103,8 @@
         var self = this;
 
         var promise = self.ready().then(function() {
-            var keyPrefix = self._dbInfo.keyPrefix;
+            var dbInfo = self._dbInfo;
+            var keyPrefix = dbInfo.keyPrefix;
             var keyPrefixLength = keyPrefix.length;
             var length = localStorage.length;
 
@@ -128,7 +128,7 @@
                 // key is likely undefined and we'll pass it straight
                 // to the iterator.
                 if (value) {
-                    value = serializer.deserialize(value);
+                    value = dbInfo.serializer.deserialize(value);
                 }
 
                 value = iterator(value, key.substring(keyPrefixLength),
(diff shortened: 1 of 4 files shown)


thgreasi:
Didn't add dist files to the PR to ease merging :)

tofumatt:
Looks good to me. I'll merge this in and wait for #452 then try to get 1.3 out :smile:

af7:
@tomufatt, when do you believe you will be able to release 1.3? It seems that with 1.2.9 I have problems with `webpack` and its `uglify` plugin (the size of the project is not reduced as when I change the 'dist' to 'src' on `package.json`). Did you experience this also?

tofumatt:
Looking to do it soon, I think possibly tonight. Sorry for the lateness on this!

af7:
Also:

If 'dist' is to remain, wouldn't it be more appropriate to use the '.min' file instead?:
`  "webpack": "dist/localforage.min.js",`

(we now have `  "webpack": "dist/localforage.js",`)

thgreasi:
Yea, but what if you needed to do a debug on your build?
I guess that if you already have a build step for WebPack, it should be easy to add an Uglification step as well (in case you don't already have).
@tofumatt Your call, I literally have no problem with this.

thgreasi:
Just to note, 1.2.10 shouldn't have any problems with webpack or browserify. I've written tests for that.

tofumatt:
I prefer not pointing people to a minified version of the code; if you’re using webpack/etc. you likely have your own step for that in production, and this way on the off-chance you want to trace an error through the localforage codebase you can do so easily.

The minified version is really for people who want to download the library as a zip and use it. I’ve even given some thought to not including it at all, but that’s a separate discussion.

af7:
Ok, I understand, please have in mind my goal too: Use webpack's **uglify** to create the smallest possible (.min) output of my application (that uses localforage). I will have to test 1.3 for that and will let you know. It will be great if you can release 1.3 soon in order to try it with the 'uglify' in webpack.

af7:
More info on this:

I did what the README says (i.e. used the `require('script!localforage');` way to require localforage, without changing the `dist` to `src`) and the size of my (uglified) webpack project is **155K**, which is too large comparing to the **84K** produced when I use `import localforage from 'localforage';` and change the `dist` to `src` (which causes other artifacts though).

Please have that in mind for the webpack solution (localforage needs to be properly uglified like any other package, within our compiled bundle).

Looking forward to the release of 1.3. Thanks!

PR #461, opened 2015-10-09
Try Karma and use NPM for deps 💣

tofumatt:
Totally a work-in-progress; aims to solve problems with testing.


nolanlawson:
Is this still a WIP? I like the idea of having a `test.js` file instead of multiple HTML pages.

I'd also probably recommend [Zuul](https://github.com/defunctzombie/zuul) instead of Karma since it offers a very low-friction way of running the tests in Saucelabs. Although karma also has [karma-sauce-launcher](https://github.com/karma-runner/karma-sauce-launcher).

.travis.yml
```diff
@@ -1,6 +1,6 @@
 before_install:
-- npm install -g bower
-- bower install
+  - "export DISPLAY=:99.0"
+  - "sh -e /etc/init.d/xvfb start"
 language: node_js
 node_js:
 - '0.10'
```
nolanlawson: might be a good time to upgrade to node 5

tofumatt: Hah, yeah ^_^

tofumatt:
Still a bit of a WIP but it's close. Zuul seems neat; I'll give it a look!

nolanlawson:
Yeah, it's super easy to set up; you just define a `.zuul.yml` and then let Zuul run your Mocha tests in Saucelabs (example: [blob-util](https://github.com/nolanlawson/blob-util)).

That said, it looks like Karma solves the same set of problems, so it might not be worth rewriting this PR.

tofumatt:
This ends up drastically changing our build process. I was able to avoid grunt entirely, but I'm not sure the built localForage is as bulletproof.

There are also issues running certain tests in karma. I'm thinking I might do things like browserify/webpack tests in PhantomJS still, but do the unit tests in karma.

I think I have a way to include the ServiceWorker tests but I haven't finished it yet.

@thgreasi could you take a quick look and tell me if this is okay, aside from the missing tests which I'll add back before moving ahead? I mostly want to know if this simplified way of building is okay (or if I should move some conf over).

thgreasi:
I will take a look at it during the weekend :)

> --
*Thodoris Greasidis*
Computer, Networking &
Communications Engineer

thgreasi:
@tofumatt I like the new structure, it looks simpler.
I guess that `npm run-script test` is the new way to test localforage, am I correct (`grunt test` any more fails since it can't find the `saucelabs-browsers.js` file)? Will we be bringing back the rest of the tests (like `test.serviceworkers.js` and `test.nodriver.js`)?

thgreasi:
@tofumatt This review took me longer that I thought.
I really like the ES6 syntax in the tests (`() => {}`) but what is really awesome is that we no longer need to use `done()`. Returning a Promise seems so natural! :+1:

## Tests
After a comparison I found that some test "suits" are currently missing:
* test.callwhenready.js
* test.iframes.js
* test.nodriver.js
  * The same approuch with what you did in test.faultydriver.js would be perfect. Really liked the `before() & after()` in there.
* test.serviceworkers.js
* integration tests (perhaps in a seperate longer-running grunt task)
   * browserify
   * webpack
   * requirejs
   * using `localforage.min.js` (perhaps we can omit this)
   * should we also add tests for `localforage.nopromises.js` after #562

Are there any plans to migrate them as well?

## Gruntfile.js
I guess we can remove the mocha unit tests that still refer the old test urls like:
`http://localhost:9999/test/test.main.html`

## Weird
On my laptop I get 4 FAILED, 475 SUCCESS and 65 SLOW tests  ([logs here](http://pastebin.com/iPHfNcZw)).
Since in Travis everything works, this might be related to the removal of the timeout increases that we used to have in some tests and/or my laptop being lazy.

PR #470, opened 2015-10-23
Check if self is present (not in webworker).

### LINKED ISSUE #469: self is not defined
See https://github.com/mozilla/localForage/blob/master/src/localforage.js#L424

```
call(typeof window !== 'undefined' ? window : self);
```

The `self` is undefined. Although I am uncertain what it is supposed to be. It looks like the passed in variable isn't used. The self might have been added in a pull request adding service worker support.

This error prevents localforage being upstream with webpack.

https://github.com/mozilla/localForage/commit/66ed3aeb261886c60f7805a1ccdbbb59d675eae0

samkelleher:
Permits webpack to use the package since self will only resolve is needed. If neither self or window are present, and empty object is used.

Fixes #469

### DIFF
--- a/src/localforage.js
+++ b/src/localforage.js
@@ -426,5 +426,5 @@ var localForage = (function(globalObject) {
     // The actual localForage object that we expose as a module or via a
     // global. It's extended by pulling in one of our other libraries.
     return new LocalForage();
-})(typeof window !== 'undefined' ? window : self);
+})(typeof window !== 'undefined' ? window : (typeof self !== 'undefined' ? self : {}));
 export default localForage;


thgreasi:
This change might need to be applied to all the `src` files of localforage.

thgreasi:
We have webpack [repacking](https://github.com/mozilla/localForage/blob/eb9d5b1a9f7416401a620fd6c307d45daae1f690/Gruntfile.js#L212-L219) [tests](https://github.com/mozilla/localForage/blob/master/test/test.webpack.html), any chance you can demonstrate your error in there

samkelleher:
Once I had applied just this one change, webpack operation resumed, and localforage continued to operate normally in the browser. I guess that when webpack evaluates the script, it has neither window or self. So confident to say this change alone should be okay.

mgcrea:
:+1: Needed! Thanks!

thgreasi:
I guess we can merge this now that its confirmed to be required two
different build configurations. :+1:
I still can't understand the problem and the fact that this change isn't
required for all occurrences of `self` puzzles me even more.

A test case would be a nice to have, but as far as it concerns me, this is
a gogo for merging even without it.

-- 
*Thodoris Greasidis*
Computer, Networking &
Communications Engineer

samkelleher:
The instances of `self` within the closure are just reflections of `this`. When the closure is executed with `window` or `self`, they are set as `this` depending if the code is being executed inside a webworker or not.

The only reason I can think of as to why it doesn't work for us but still passes the test is that the version of webpack or babel (newer than in localforage) is now (rightly so I think) sensitive to `self` being undefined without checking first.

The repack entry point used in the test uses a require() call vs an module import which is what we're using, so maybe it has something to do with that.

thgreasi:
Oh ok. We should then add a es6/babel + webpack test case to replicate this!

On Sat, Nov 28, 2015, 21:50 Sam Kelleher <notifications@github.com> wrote:

> The instances of self within the closure are just reflections of this.
> When the closure is executed with window or self, they are set as this
> depending if the code is being executed inside a webworker or not.
>
> The only reason I can think of as to why it doesn't work for us but still
> passes the test is that the version of webpack or babel (newer than in
> localforage) is now (rightly so I think) sensitive to self being
> undefined without checking first.
>
> The repack entry point used in the test uses a require() call vs an module
> import which is what we're using, so maybe it has something to do with that.
>
> —
> <https://github.com/mozilla/localForage/pull/470#issuecomment-160332001>.
>
-- 
*Thodoris Greasidis*
Computer, Networking &
Communications Engineer

jondkoon:
Would love to see this merged and published. Just ran into this issue when trying to upgrade from 1.2.10 to 1.3.1 and making use of the webpack compatibility.

thgreasi:
Any change you can rebase this PR to the latest commits?

samkelleher:
@thgreasi PR rebased. 👻

tofumatt:
I'm good to merge this, but is there a test we could add to make sure it stays working?

thgreasi:
I can't think of a way to test this, except by replicating the build setup
that was originally causing the issue.
Also, I would prefer it, if the same code was used in the rest files as
well (serializer & drivers) for consistency.

> --
*Thodoris Greasidis*
Computer, Networking &
Communications Engineer

tofumatt:
Hmm, good point. And yeah, using this everywhere for consistency is a good
call. I can eventually do that if no one gets to it, but otherwise once
that's done we should merge it :-)
- tofumatt

okcoker:
:+1:

okcoker:
Submitted #531 in hope that this was the last step in getting this merged!

Issue #471, opened 2015-10-27
`asyncStorage` driver fails to init within EmberJS


sebweaver:
Actually the culprit is this block within the `_initStorage()`:

```javascript
for (var k in forages) {
  ...
}
```

In EmberJS, arrays are specialized, and this loop iterates over the special key `_super` which fails further.

I will create a PR fixing this by a more generic iteration:

```javascript
for (var k = 0; k < forages.length; k++) {
  ...
}
```

PR #472, opened 2015-10-27
Fix `asyncStorage` driver initialization within EmberJS - issue #471

sebweaver:


tofumatt:
Right on. Thanks! :+1: 

Any chance you could add a simple test for this (override array or something?) so we can make sure we don't undo it?

sebweaver:
Hmm, I don't really see how to test this the right way. It's not an API, not a use case, it's just a code pattern which was not relevant for what it is intended to do.

I mean, `for (x in y)` pattern should be reserved for iterating over attributes of objects, whereas `for (i = 0; i < a.length; i++)` pattern is more relevant for iterating over indices of plain arrays. I made this mistake when submitting #452, but to my mind, it's not really necessary to test this. Don't you think?

If you think it's worth it, let me know how and I could implement it and add it to the PR.

tofumatt:
I won’t get hung up on it then; it passes our tests and is a good fix!

Issue #476, opened 2015-11-20
Check instanceof Blob before check Blob support at IndexedDB


spirinvladimir:
Check instanceof Blob before check Blob support at IndexedDB.
https://github.com/mozilla/localForage/pull/477

tofumatt:
Closed in #477.

PR #477, opened 2015-11-20
check first instanceof Blob

spirinvladimir:
This changes are only for IndexedDB.

Main goal of PR is check type of value before check browser's blob support.

**Before changes**:
First check browser's blob support. Then if needed check value type.
**After changes**:
First check value type. Then if needed check browser's blob support.

Browser's blob support check make fake write blob to IndexedDB. And store result of check in private variable [supportsBlobs](https://github.com/mozilla/localForage/blob/master/src/drivers/indexeddb.js#L18)
Second browser's blob support check will use ```supportsBlobs``` without fake write blob to IndexedDB. 

I add 3 new tests:
* Save non blob value should not trigger browser's blob support check with fake save blob.
* Save blob value should trigger browser's blob support check with fake save blob.
* Second save blob value should use variable ```supportsBlobs``` and do not check fake save blob.

Blob support check functions and variable ```supportsBlobs``` are private. That is why I use mock db.transaction. After tests I return db.transaction to original state.

Issue is here:
https://github.com/mozilla/localForage/issues/476


dist/localforage.nopromises.js
```diff
@@ -718,7 +718,7 @@ return /******/ (function(modules) { // webpackBootstrap
 	            dbInfo.db = dbContext.db = db;
 	            self._dbInfo = dbInfo;
 	            // Share the final connection amongst related localForages.
-	            for (var k in forages) {
+	            for (var k = 0; k < forages.length; k++) {
```
tofumatt: Interesting that this change was made by the compiler.

tofumatt:
Looks good; can you add a test for this?

dist/localforage.nopromises.js
sebweaver: It's related to https://github.com/mozilla/localForage/commit/47aaaa82c7c86416e9e000973c5b105ca868e8cd

tofumatt: Oh right, forgot to add the dust files into that commit, oops!

spirinvladimir:
I'v added 3 tests.

test/test.api.js
```diff
@@ -223,6 +223,68 @@ DRIVERS.forEach(function(driverName) {
             });
         });
 
+        if (driverName === localforage.INDEXEDDB) {
```
tofumatt: I don't see any `expect` calls in these tests so I don't see what they're testing for. If you want `called` to be more than one you'll want to test for its value with `expect`, which I think is what you're going for.

tofumatt: Thanks for adding the `expect` calls, but I'm curious: what _exactly_ is this testing for? I'm a bit confused and the PR doesn't have a description nor is this related to an issue.

What bug does this prevent and what are these tests doing?

spirinvladimir:
Sorry for PR without description. I've add it now.

tofumatt:
Aha, cool!

That makes sense, awesome. I'll review this quickly in the evening then, but it looks good. Should be able to merge it tonight, sorry for the back-and-forth! :smile: 

Thanks! :smile:

PR #486, opened 2015-12-12
feat(src): change the sources to be babel non-loose compatible

### LINKED ISSUE #467: Move import/export statements to top level
As mentioned in #466 and possibly #465, we have `import`/`export` statements outside the top-level which requires a config option.

We should see about moving them out so users don't need to change their babel configs just because of localForage :smile:

thgreasi:
Resolves #467.
Still uses babel's loose mode, for our own compilations, in order to produce smaller and more readable code.


tofumatt:
Sweet! Thanks :+1:

Issue #482, opened 2015-12-05
Error in Safari when running in web worker


snorpey:
Hi there.
Thank you for making this great library!

I'm trying to use localForage in a web worker and noticed an error in Safari:

[``localforage.nopromises.js:105 TypeError: undefined is not an object (evaluating 'self.indexedDB')``](https://github.com/mozilla/localForage/blob/master/dist/localforage.nopromises.js#L105)

I'm using Safari version ``9.0.1 (10601.2.7.2)``.

You can find a gist of the code I used [here](https://gist.github.com/snorpey/f7ddb3b8cfa928ce6dcd) or download it [here](https://gist.github.com/snorpey/f7ddb3b8cfa928ce6dcd/archive/1d0233b6c532d5073dac05924554aa732e95406c.zip)

To reproduce the error, download all files of the gist and then open ``index.html`` in Safari.

jamessharp:
I've also bumped into this. I think it comes down to a safari bug but I've got a workaround for it in #493.

PR #493, opened 2016-01-07
Error in safari when running as web worker

jamessharp:
As described in #482 there's a problem running local forage in a safari web worker.

Unfortunately there seems to be a Safaribug (at least in Safari 9) where running something like the following as a web worker results in `this` being undefined.

```javascript
'use strict';
(function() {
  self.postMessage(typeof this); // `this` is undefined
}).call(self);
```

It only happens in strict mode (i.e. remove the 'use strict' and everything is OK). This PR works around the issue by passing in the global variable directly (rather than using `call`)


thgreasi:
Babel will add the `'use strict';` statement anyway (since ES6 modules run in strict mode) so :+1: for your IIFE transformation :smile:

Issue #494, opened 2016-01-09
TypeScript typings, issue with 'default'


dsebastien:
Hello,

I'm currently trying to use localForage in a TypeScript project. As I've explained in #450 , the way the localForage typings file is currently implemented make it difficult for me to get auto completion.

I can get my IDE to work fine by defining the ambient module using the following syntax:

```
declare module "localforage" {
	export var localForage: LocalForage;
}
```

Instead of:

```
declare module "localforage" {
    var localforage: LocalForage;
    export default localforage;
}
```

Is the `default` export really necessary or could would you consider changing this?

thgreasi:
I think that it is, since the default instance has to be attached to the
window object automatically when not loaded with a more modern way.

> --
*Thodoris Greasidis*
Computer, Networking &
Communications Engineer

dsebastien:
What do you mean by "modern way"?
Currently, I'm using JSPM & SystemJS, thus I would assume that I use a modern way :smile: 

Here's the output of my tests:

1) Without 'default' in the typings as I proposed:
```
declare module "localforage" {
	export var localForage: LocalForage;
}

and the following import:
```
import {localForage} from "localforage";
```

tsc outputs:
```
var localforage_1 = require("localforage");
```

2) With 'default' in the typings:
```
declare module "localforage" {
	var localForage: LocalForage;
	export default localForage;
}
```

and the following import:
```
import {default as localForage} from "localforage";
```

tsc outputs:
```
var localforage_1 = require("localforage");
```

Thus I don't think 'default' in the d.ts file make any difference apart for the way we can write the import (?)

thgreasi:
I might be mistaken. Was referring to the default keyword in the .js file not the .ts. Sorry.
If it still matters, as "non-modern/traditional way" I was referring to including localforage in a script tag.

dsebastien:
Script what? :D
Yep I'm sure you're right regarding localForage's js code, it certainly needs to have default in the exports.

Okay, I'll open a PR on DefinitelyTyped to propose the typings change and another one against this repository to update the readme.

dsebastien:
Actually I overlooked something..

In the generated code, the require statement is as I've described above, but the usages of the associated variables are:

1) Without 'default' in the typings as I proposed: `localforage_1.localforage.setItem("key", "cool");`
2) With 'default' in the typings: `localforage_1.default.setItem("key", "cool");`

[...]

PR #495, opened 2016-01-10
Updated README regarding TypeScript usage

dsebastien:
See discussion in #494

### DIFF
--- a/README.md
+++ b/README.md
@@ -208,8 +208,12 @@ installed (eg. `npm install --save-dev babel-plugin-system-import-transformer`).
 To import localForage in TypeScript:
 
 ```javascript
-import {default as localforage} from "localforage"
+const localForage:LocalForage = require("localforage");
 ```
+
+Note that the ES6 style import is not supported for our module type. Check out the following to know why:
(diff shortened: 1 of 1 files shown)


tofumatt:
I don't know really anything about typescript, but why declare it as a constant?

dsebastien:
I feels safer to me, although I'm no TypeScript guru either ;-))

tofumatt:
If there's no clear reason to do it, let's not make it a constant without understanding the ramifications. But after that fix I'd say we're good to merge, thanks :-)

- tofumatt

dsebastien:
let and const have been supported in TypeScript since version 1.5 and can be transpiled down to ES3 & ES5, thus logically this should not have any negative effects.

AFAIK, using let and const is the way to go; var should never be used anymore and here, the choice between let and const depends on whether you want something mutable or not; for an import I'd say that const is probably a safer recommendation.

If you want maybe I can add this explanation to the readme?

tofumatt:
Okay, that makes sense.

The readme is busy enough so it's okay, but I'll keep it in mind when I get the chance to fix up the docs (an idea I have for the future).

- tofumatt (Sent from mobile)

Issue #492, opened 2015-12-29
Error on indexDB first startup


TCMiranda:
Hello guys,

I'am having the same problem as this guy:
http://stackoverflow.com/questions/26843311/after-clearing-cache-and-localforage-indexeddb-getting-an-error-on-loading-the

But I found no answer.

After clearing the indexedDB storage, and starting the app normally, it fails with the following:
![captura de tela de 2015-12-29 18 32 12](https://cloud.githubusercontent.com/assets/10077439/12041664/82142af4-ae5a-11e5-993d-f350540dfc8d.png)

The line referenced is:
![captura de tela de 2015-12-29 18 33 40](https://cloud.githubusercontent.com/assets/10077439/12041682/b43f31b8-ae5a-11e5-8be9-1db8de640829.png)

My code is only opening the instance and trying to get an item:
![captura de tela de 2015-12-29 18 35 21](https://cloud.githubusercontent.com/assets/10077439/12041713/fd367f2a-ae5a-11e5-98cf-4004ee2d6095.png)
And then...
![captura de tela de 2015-12-29 18 35 38](https://cloud.githubusercontent.com/assets/10077439/12041715/01f3757c-ae5b-11e5-8828-cba0f0984809.png)

After a refresh, it runs normally.

I am using localforage 1.3.1 no promises.

Thanks!!

TCMiranda:
Anyone else experiencing this issue?

Thanks!

sebweaver:
@TCMiranda I can reproduce this error too with Chrome.

In FF, the error message is a bit different:

    InvalidStateError: A mutation operation was attempted on a database that did not allow mutations.

But the reason is the same, and I think I have figured out it.

I will commit a PR very soon to expose the problem and fix it.

siddo420:
I am getting the same error in FF. Whats the status of this issue?

```
InvalidStateError: A mutation operation was attempted on a database that did not allow mutations.
var transaction = dbInfo.db.transaction(dbInfo.storeName, 'readwrite');
localforage.js (line 1608)
```

line 1608:
```
	                return value;
	            }).then(function (value) {
	                var transaction = dbInfo.db.transaction(dbInfo.storeName, 'readwrite');  //1608
	                var store = transaction.objectStore(dbInfo.storeName);
```

tofumatt:
The PR has been merged but I need to release a new version of localForage. I'll do that today and report back when it's available.

tofumatt:
[...]

PR #497, opened 2016-01-12
Fix issue #492 (instance concurrency on same database)

sebweaver:
Specialize the default `ready()` function by making it dependent
on the current database operations. Thus, the driver will be actually
ready when it's been initialized (default) *and* there are no pending
operations on the database (initiated by some other instances).

### DIFF
--- a/src/drivers/indexeddb.js
+++ b/src/drivers/indexeddb.js
@@ -168,6 +168,23 @@ var asyncStorage = (function() {
         return value && value.__local_forage_encoded_blob;
     }
 
+    // Specialize the default `ready()` function by making it dependant
+    // on the current database operations. Thus, the driver will be actually
+    // ready when it's been initialized (default) *and* there are no pending
+    // operations on the database (initiated by some other instances).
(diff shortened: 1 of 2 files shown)


tofumatt:
I can't quite tell if the existing tests cover this change or if we should add another to make sure we don't lose this feature and it's working as intended.

But :+1:

sebweaver:
I forced the database to be [deleted](https://github.com/sebweaver/localForage/blob/issue-492-instance-concurrency-on-same-db/test/test.api.js#L853-L873) before the existing tests on multiple instances.

And the latter failed until I added the new implementation.

So I think this is sufficient, but If you have a different test case in mind, I could add it to this PR.

tofumatt:
Cool, okay, that makes sense :+1: wfm

sebweaver:
I think you might had right about the test coverage, since this PR created new issues...

Anyway, I created a new one here: https://github.com/mozilla/localForage/pull/507 (with a more dedicated new test)

Issue #496, opened 2016-01-11
incompatible with projects that use Babel 6, as "loose" is now unknown option


jokeyrhyme:
I have a project that uses Babel 6. When I try to `require('localforage')`, I get the following babel compile error:

> ERROR in ./~/localforage/dist/localforage.js
> Module build failed: ReferenceError: [BABEL] redacted/node_modules/localforage/dist/localforage.js: Unknown option: redacted/node_modules/localforage/.babelrc.loose

The "loose" global option seems to be gone now: https://babeljs.io/docs/usage/options/

It looks like this is now supposed to be a per-plugin option, e.g: https://github.com/babel/babel/blob/6.4-stable/packages/babel-plugin-transform-es2015-for-of/test/fixtures/loose/options.json

This is loosely (pardon the pun) related to the following PRs:

- https://github.com/mozilla/localForage/pull/486
- https://github.com/mozilla/localForage/pull/390

Besides updating localForage to use Babel 6 as well, I'm not sure what other actions can be taken to resolve this.

jokeyrhyme:
I am using webpack with babel-loader, and I've changed the way I use it, from:

`{ test: /\.jsx?$/, loader: 'babel-loader' }`

to:

`{ exclude: /localforage/, test: /\.jsx?$/, loader: 'babel-loader' }`

So my project can compile with Babel 6, without ever detecting the Babel 5 configuration that is supplied with the localforage npm module.

thgreasi:
I already have a dev branch to upgrade to babel 6. I will continue my work
on it as soon as I get some free time.

> --
*Thodoris Greasidis*
Computer, Networking &
Communications Engineer

thgreasi:
May we can just remove the configuration from the npm package?
-- 
*Thodoris Greasidis*
Computer, Networking &
Communications Engineer

jokeyrhyme:
The problem with removing the .babelrc file, is that babel will refuse to touch any of the files in the package. Which is fine for babel-using consumers who do `import localforage from 'localforage';`, but will break any consumers that do (for example) `import localforage from 'localforage/src/localforage';`.

This is only an interim problem, as you say, because:

1. babel 6 support is coming soon, anyway, and
2. ES2015 source files within an npm package will eventually be able to run on V8 / Node.js without babel

thgreasi:
First of all using the src files is not officially supported by this repo
as far as I know but we are not trying to make it impossible either.
Since we can't support two different versions of babel at the same time
[...]

PR #498, opened 2016-01-12
move .babelrc closer to files it governs

jokeyrhyme:
- may be an interim solution to #496 
- affects projects that consume the default package "main", or other files in the "dist" directory
- for such projects that use babel themselves, the .babelrc for localforage will no longer be in scope


tofumatt:
I know @thgreasi said he was working on a fix for this, so an interim fix might be a bit hasty.

How does this affect current users exactly? I don't understand the bit about users using `package.json`'s `main`--will they have to change configs?

In that case it's just a big enough/breaking change that will require a bump to `1.4`.

jokeyrhyme:
There's no breaking change.

> affects projects that consume the default package "main"

By this I mean projects that just `import localforage from 'localforage';` (the designated "main" file in package.json).

>  or other files in the "dist" directory

By this I mean projects that `import localforage from 'localforage/dist/localforage.nopromises';` (for example)

No consuming project has to make any changes as a consequence of this. In fact, this allows more consuming projects to build correctly with babel.

jokeyrhyme:
The only downside (if any) is that all files in this project that need babel MUST be located within the "src" directory. As this is already the case, and is likely to be the case for the foreseeable future, it doesn't seem like such bad limitation.

tofumatt:
Ah, cool. Sorry I misunderstood.

Looks good then, I’ll merge this in a few ^_^

tofumatt:
Actually, this isn't what we'd want, as if we wanted to start using ES6 for tests (which I personally would), it wouldn't be covered. Other projects I work on keep the `.babelrc` in the root of the project, that seems correct. So this seems quite temporary, I'm torn :-/

jokeyrhyme:
Yeah, it's definitely a weird compromise. Thought I'd run this past you anyhow and see what people thought. Having a second .babelrc in the "test" directory also seems weird.

As you say, this is of limited benefit once babel 6 is supported.

jokeyrhyme:
Actually, even once babel 6 is supported, there'll still be some theoretical value here.

In babel 6, the "loose" option was dropped (well, moved) which made old .babelrc files throw errors. If babel 7 (for example) similarly breaks when it reads .babelrc files for babel 6, then we'll just run into the same problem next time.

Ideally, we want the files that are babel'ed here already (those in "dist") to have nothing to do with babel.

thgreasi:
:+1:
Didn't know you could do such a thing!
If it works as I think then it might be even better than removing the
.babelrc file from the npm package. Since package.json's main targets the
dist files, trying to force everyone to use a single import method, this
could conceal the use of babel, except from those that manually target the
src files.

Ps: I guess that this is not affecting the build at all?

Ps2: I won't be able to resume the upgrade to babel 6 until early next
month :(

> --
*Thodoris Greasidis*
Computer, Networking &
Communications Engineer

tofumatt:
Looks like the issue this meant to fix has been resolved, this commit is out of date, and I think we want to close this.

Thanks for the patch but I think this is fixed now? Please let me know if I'm wrong to close.

jokeyrhyme:
If we're using the latest babel now, then there's no need for this.

PR #505, opened 2016-01-19
Fixed wrongly handled errors

rromanw:
My previous commit was partially wrong, I wasn't able to reproduce the issue with firefox.
The submitted error handling therefore meaningless and it seems it is buggy too.


tofumatt:
Fair enough, reverting it!

PR #507, opened 2016-01-21
Actual fix for issue #492 (instance concurrency on same database)

sebweaver:
The previous PR was flawed, whereas the tests passed. The latter might have not reflected the actual error cases (perhaps related to the `before` and `beforeEach` promises).

This updated PR:
- adds a new dedicated test, with multiple operations on 3 stores sharing the same database.
- revamps the previous flawed strategy in order to fix this issue once and for all (I hope).


tofumatt:
Looks good to me! Thanks for your work on this :+1:

PR #510, opened 2016-01-26
fix for bug where we were going into an infinite loop on fail

marcucio:
fix for bug where we were going into an infinite loop if the open database failed (saw this on BB10 cordova app)

I don't think we should call _initStorage again since it just failed


tofumatt:
Odd that it goes into an infinite loop when we explicitly choose localStorage, which should then just reject again and then stop on failure, right?

tofumatt:
It would be handy to include a test that checks for this if we decide to merge it in so we don't revert it in the future. I saw you had an example in #511 -- could you add that to the tests for this PR?

marcucio:
ok, I'll see if I can come up with a test for this, it might be a few days since I'm trying to get a release out for one of my apps.

marcucio:
@thgreasi @tofumatt 
I integrated test cases from #515 into this pull request, everything seems good to go. One question, should I also check in the dist files?

thgreasi:
Add the dist files as well.
How about pulling or rebasing the test case instead of copying?

marcucio:
sorry, I'm not sure how to merge a pull request from this fork into my fork, I will have to look into how to do that.

I just checked in dist files so that should be good to go too.

marcucio:
@thgreasi @tofumatt I think this should be all set to merge, I merged the #515 in to this request

thgreasi:
Great :+1:
I guess that this should be released with a new minor version change, since there might be people that relied on the hidden websql->localStorage fallback mechanism.

thgreasi:
Also :+1: for adding the dist files as a separate commit!

tofumatt:
Looks good, I may try to squash a few of these commits before I merge and it’s getting late here, so I might not get to this until tomorrow, but rest assured I’m on it! Thanks!

tofumatt:
Merged in #517 with some tidying.

Issue #511, opened 2016-01-26
Infinite loop on driver open fail


marcucio:
fix for bug where we were going into an infinite loop if the open database failed (saw this on BB10 cordova app)

I don't think we should call _initStorage again since it just failed

fixed in pull request: #510

thgreasi:
Perhaps we could remove this fallback and just reject the promise. The mechanism in the main localforage.js module can handle the fallback to the next preferred driver. I had seen that earlier, before the es6 refactor and started a discussion about it in an other issue, but can't find it right now. This was probably used before the fallback-on-fail mechanism was introduced, to prevent some missconfigurations.

marcucio:
@thgreasi so your saying we might not have to do self.setDriver(self.LOCALSTORAGE), it looks like indexdb does it this way (just resolves it):

```
        function ignoreErrors() {
            // Don't handle errors here,
            // just makes sure related localForages aren't pending.
            return Promise.resolve();
        }

        for (var j = 0; j < dbContext.forages.length; j++) {
            var forage = dbContext.forages[j];
            if (forage !== self) { // Don't wait for itself...
                initPromises.push(forage._initReady().catch(ignoreErrors));
            }
        }

```

thgreasi:
Exactly. I think that catch should just do a reject().
Any chance you can provide a test case or a code example with the infinite
loop?

> --
*Thodoris Greasidis*
Computer, Networking &
Communications Engineer

marcucio:
yes, it is easy to repro:
in websql.js replace:
```
dbInfo.db = openDatabase(dbInfo.name, String(dbInfo.version), dbInfo.description, dbInfo.size);
``
with:

```
throw new Error('quota exceded');
```

results in this:
```
Running "mocha:unit" (mocha) task
Testing: http://localhost:9999/test/test.component.html

  
  ․
Running PhantomJS...ERROR
[...]

PR #515, opened 2016-01-28
test(websql): test the case that openDriver fails

thgreasi:
Related test cases for #510 and #511.
The required change to make the test pass is to reject in case that openDatabase fails (instead of setting the localforage driver).
```js
try {
    dbInfo.db = openDatabase(dbInfo.name, String(dbInfo.version),
                             dbInfo.description, dbInfo.size);
} catch (e) {
    reject(e);
}
```
The required src change is left for @marcucio to implement since he initially found the issue.


tofumatt:
Merged in #517 with some tidying.

PR #517, opened 2016-02-03
fix for bug where we were going into an infinite loop if the open dat…

tofumatt:
Condensed for atomic commit.


tofumatt:
Oops, didn't mean to squash the tests... gonna fix that.

PR #528, opened 2016-02-20
chore: update to babel 6

### LINKED ISSUE #496: incompatible with projects that use Babel 6, as "loose" is now unknown option

thgreasi:
Resolves #496 
Also disables the Component test (which is deprecated anyway) due a change in the UMD pattern that babel6 uses.


thgreasi:
We could also remove:
* the commented code in `Gruntfile.js`
* the file with the test cases
* the npm package from `package.json`

tofumatt:
I'm for all that, we should do it here.

- tofumatt

thgreasi:
@tofumatt Just pushed the commit that completely drops Component support.

tofumatt:
Sweet, this looks great. Thanks!

PR #529, opened 2016-02-27
feat(localforage): check if a callback is defined before passing to then()

### LINKED ISSUE #499: Warning: .then() only accepts functions but was passed: [object Undefined], [object Undefined]
We've been getting this warning since we started using LocalForage. We are using Bluebird.js

Here is the exact warning:

```
Warning: .then() only accepts functions but was passed: [object Undefined], [object Undefined]
    at LocalForage.setDriver (http://localhost:8000/assets/js/vision.bundle.js:75077:31)
```

The exact code where it fails was pointing to this code in localforage.js
```
this._driverSet.then(callback, errorCallback);
```

thgreasi:
Properly resolves #499


tofumatt:
Cool, thanks! I'll cut a release later, right now I'm on my phone and technically on holiday so I'm being lazy ;-)

PR #538, opened 2016-03-06
chore(localforage): promise -> lie, use ES6 modules

nolanlawson:
This is kind of a huge refactor on my part, so it's incumbent upon me to justify it.

Goals:

1. Reduce `dist` size. No huge wins here, but the min+gz size did go from 8197 to 7676, so it's at least something.
2. Use `lie` in favor of `promise` (since it's smaller, hence the gains above)
3. Use ES6 modules instead of SystemJS (easier to bundle, more modern)
4. Use Browserify+Rollupify as the primary module bundler (Babel during tests). Rollup reduces bundle size by hoisting all `import`ed dependencies into a single scope.
5. Move common code into shared ES6 modules instead of repeating ourselves. Using Rollup means this is essentially costless; the browserify bundle basically ends up as one big `index.js`.

I'll be honest; I was hoping that the gains would be much larger, since `lie` is so much smaller than `promise`. When I swap in `promise` for `lie` with my other changes, it goes from 7676 to 8892, so it's indeed larger, just not as much as I expected. (That 8892 number also indicates I added some cruft with my non-promise changes, but TBH I can't figure out where exactly those bytes came from.)

So I suppose I can only justify this PR based on the cleaner `src` format (more modularity, modern build system with ES6 modules). I also migrated from building for production with Webpack to using Browserify+Rollupify instead, but that's just because of my personal preference for Browserify and the fact that I know how to easily get it to do what I want it to do. As long as it's building UMD, it shouldn't matter one way or the other.

Tests are failing in Chrome, but this seems to be unrelated to this PR since master is failing as well.


Gruntfile.js
```diff
@@ -53,6 +53,45 @@ module.exports = exports = function(grunt) {
...
+                options: {
+                    browserifyOptions: {
+                        standalone: 'localforage'
+                    },
+                    transform: ['rollupify', 'babelify'],
+                    plugin: ['bundle-collapser/plugin'],
+                    exclude: ['lie']
+                }
+            }
+        },
```
nolanlawson: the "no promises" build is now just a browserify builde with `exclude: ['lie']`

Gruntfile.js
```diff
@@ -53,6 +53,45 @@ module.exports = exports = function(grunt) {
...
+                cmd: './bin/derequire.js',
+                args: [
+                    'dist/localforage.js'
+                ]
+            },
+            derequire_no_promises: {
+                cmd: './bin/derequire.js',
+                args: [
+                    'dist/localforage.nopromises.js'
+                ]
```
nolanlawson: derequire is needed in order to have the `dist` bundle compatible with browserify itself (otherwise it complains when it sees the keyword `require`)

src/drivers/indexeddb.js
```diff
@@ -1,766 +1,722 @@
+import createBlob from '../utils/createBlob';
+import idb from '../utils/idb';
+import Promise from '../utils/promise';
+import executeCallback from '../utils/executeCallback';
```
nolanlawson: Notice all the ES6 modules; this is gonna be a theme of my PR. Where possible I moved stuff to separate functions and then just `import`ed them.

src/utils/idb.js
```diff
@@ -0,0 +1,21 @@
...
+    if (typeof mozIndexedDB !== 'undefined') {
+        return mozIndexedDB;
+    }
+    if (typeof OIndexedDB !== 'undefined') {
+        return OIndexedDB;
+    }
+    if (typeof msIndexedDB !== 'undefined') {
+        return msIndexedDB;
+    }
+}
```
nolanlawson: Another change I made in a lot of places is to avoid checking the `global` object (e.g. `global.indexedDB`) and instead just check `typeof indexedDB !== 'undefined'`. I've found that this is less faily for weird environment like NodeWebkit where you can't be sure what the global object is even called. It should also make the code more portable if you do want to switch between Webpack, Browserify, SystemJS, etc.

nolanlawson:
Tests are failing due to some kind of problem with PhantomJS/promises in the browser tests. Kinda hard to debug since PhantomJS is so old.

nolanlawson:
OK, finally got all the tests passing (`npm t` works, using Phantom 1.x). The only thing I had to remove were the tests for the "unbundled" API, but the code comments said that that hasn't been supported since 1.3 anyway. I also think it would be pretty difficult to unbundle this if we switch to ES6 modules.

test/test.faultydriver.html
```diff
@@ -25,6 +25,9 @@
     <!-- Modernizr  -->
     <script src="/bower_components/modernizr/modernizr.js"></script>
 
+    <!-- Promise polyfill -->
+    <script src="/bower_components/es6-promise/promise.js"></script>
+
```
nolanlawson: I couldn't figure out how to get `window.require('promise')` to work as it was working in the old tests, so I figured it was easier to just drop in a promise polyfill for the tests. Notice I only inserted them where the tests themselves relied on `window.Promise`; `lie` itself only relies on `window.Promise` in `utils/promise.js`, where it prefers the global Promise and then falls back to `lie` (or falls back to nothing, in the case of `nopromises`).

nolanlawson:
I have some other ideas for reducing build size (e.g. removing the custom classes and using oldschool ES5 instead; Babel adds a lot of bloat), but I'll save that for a later day, since this PR is already massive. :stuck_out_tongue:

nolanlawson:
Interesting - it looks like the "additional cruft" is a GZip artifact. If you discount GZIp, then this PR reduces filesize across the board, including `nopromises` where the `lie`/`promise` distinction wont' make a difference.

My hunch is that it's because of Rollup hoisting everything up to one scope, compared to the repeated code across the three adapters. GZip is able to compress that better if it's not in a single scope, because then the uglify variables end up being exactly the same between all adapters. I think we can fix this in the future by just moving more logic out of those adapters and into util functions.

thgreasi:
I think that webpack was giving better file size results than browserify
when we were deciding for the bundling system, but that might have changed
now. Also webpack looked like a better match for developing libraries,
since it doesn't require the extra derequire step.
I could give a try to this refactor+webpack in the weekend and possibly try
moving rollup as a babel plugin.

> --
*Thodoris Greasidis*
Computer, Networking &
Communications Engineer

nolanlawson:
I don't have a preference one way or the other for Browserify/Webpack; you may indeed be able to get a smaller size, because I believe Browserify's header is a bit bigger than Webpack's.

There is indeed a Babel Rollup plugin, so you could give that a shot: https://github.com/rollup/rollup-plugin-babel. However note that you will probably have to run Rolup as a pre-Webpack step, because this is a _Babel plugin for Rollup_, not a Rollup plugin for Babel.

nolanlawson:
Looks like another option is to just use Rollup by itself. Apparently it has enough plugins that this can be done without Webpack/Browserify.

nolanlawson:
OK, I made an effort to use pure Rollup instead of either Webpack or Browserify, but somehow the UMD build isn't working. Also I strongly unrecommend `grunt-rollup`, since it just makes it more difficult to use Rolup's API and doesn't seem to work for multi-task. Feel free to pick up where I left off, if you want: https://github.com/nolanlawson/localForage/tree/537-rollup

NB: the Rollup build actually ends up being exactly the same size as the Rollupify+Browserify build. This starts to suggest to me that a Webpack build wouldn't be any smaller. So personally I'm happy with this PR as-is, but feel free to explore other avenues.

nolanlawson:
After some trial-and-error, I realized that `lie` actually only works with Browserify. Or rather, it only gives you the super tiny package size with Browserify, because it has a bunch of Browserify transforms that strip out the extra data.

I'm working on a Browserify-agnostic version of `lie`, but I don't have anything concrete written yet, so I'll have to re-propose this PR as-is.

nolanlawson:
OK, making progress on removing Browserify/Webpack entirely for the build (they'll only be tested in the explicit Browserify/Webpack tests). Just a few bugs remaining, but progress! (See https://github.com/rollup/rollup-plugin-node-resolve/issues/8#issuecomment-193859166 and [this branch](https://github.com/nolanlawson/localForage/tree/537-rollup)).

thgreasi:
Could you be to much to ask you to split into two commits? One with the code `reactor` and one `chore` for the build system changes.
It would be very helpful for my webpack+lie tests.
Thanks in advance. Can't wait for the weekend.

nolanlawson:
It's really the same change; I can't easily split it up into two.

The refactor involved a switch to ES6 modules. The build system change incorporated Rollup, which bundles ES6 modules.

As for your webpack tests, they're still enabled after this PR, so everything's good, right?

bin/derequire.js
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env node
+
+//
+// wrapper for derequire because there's no simple CLI. grrrr.... >:(
```
calvinmetcalf: [this is false](https://github.com/calvinmetcalf/derequire/blob/master/bin/cmd.js) :smile:

paulfalgout:
Re: the rollup plugin vs rollup vs webpack vs browserify etc.. this might help a little: https://github.com/samccone/The-cost-of-transpiling-es2015-in-2016  Arguably the tests cases is a bit simple, but I still find the results insightful.

tofumatt:
I haven't chimed in on this yet but it's a :+1: idea from me. I've been doing ES6 over the past eight months and it's been incredible so I'm for the move. The shaved bytes are cool too.

It's a clear reminder than I need to update localForage's infrastructure though, with the old PhantomJS, etc.

nolanlawson:
Testing an old PhantomJS is kinda nice in some ways, because it confirms that your implementation works against old WebSQL. :) I've seen bugs that only exist in PhantomJS 1.x but not modern Safari or Chrome WebSQL.

nolanlawson:
Rebased and fixed the conflict. Is there anything y'all would like to make this more palatable?

Gruntfile.js
```diff
@@ -53,21 +53,59 @@ module.exports = exports = function(grunt) {
...
+                    transform: ['rollupify', 'babelify'],
+                    plugin: ['bundle-collapser/plugin'],
+                    exclude: ['lie']
+                }
+            }
+        },
+        run: {
+            derequire: {
+                exec: 'derequire ' +
+                  '< dist/localforage.js > dist/localforage.tmp ' +
```
tofumatt: How well does stuff like this work on Windows?

src/drivers/indexeddb.js
```diff
@@ -1,766 +1,722 @@
...
+// Sadly we use the 404 bug to detect the FileReader bug, so if they
+// get fixed independently and released in different versions of Chrome,
+// then the bug could come back. So it's worthwhile to watch these issues:
+// 404 bug: https://code.google.com/p/chromium/issues/detail?id=447916
+// FileReader bug: https://code.google.com/p/chromium/issues/detail?id=447836
+//
+function _checkBlobSupportWithoutCaching(idb) {
+    return new Promise(function(resolve, reject) {
+        var blob = createBlob([''], {type: 'image/png'});
+        var txn = idb.transaction([DETECT_BLOB_SUPPORT_STORE], 'readwrite');
```
tofumatt: This stands for `transaction`, yeah? I'll go through and expand this, but I am afraid of variable names without vowels 😉

tofumatt:
Looking through this it seems solid. I'm gonna spend the next week working solely on localForage--one of my aims is to tidy up the tests and fix some weird fails.

I'm cool with merging this, it seems like a much cleaner `src/` with a bit of win in terms of filesize.

@thgreasi Any opposition to seeing this merged over the weekend? If you're cool with it I'm good.

Also @nolanlawson OMG thank you this is pretty amazing. You rock.

thgreasi:
@nolanlawson
* Is there a reason that `es6-promise/promise.js` is included in all test pages? We used to concatenate it (the polyfill) on the top of the dist file, is it still included to the main dist `localforage.js`? I would prefer if such a change to the test cases was not introduced, so that it's clear that we still support older browsers.
* Do we need to import Promises in `src/utils/promise.js`? Can't we just assume that the environment has been polyfilled with the required standardized APIs? (If this is just the way that `lie` prefers it, then ok, otherwise requiring polyfills had troubled us before and we removed it.

@tofumatt other than these queries this looks great to me as well. @nolanlawson :+1: 
After the first +- diff shock, it looks like that in most cases code has just moved around to make the files more logically structured.

nolanlawson:
> Do we need to import Promises in `src/utils/promise.js`?

The point of that file is to isolate our Promise usage to one single file. Then building the promises/nopromises version is as simple as adding the browserify `-x lie` flag. If `Promise` is available as a global function, we use that, otherwise we fall back to `lie` (or fall back to nothing in the case of `nopromises.js`).

> Is there a reason that `es6-promise/promise.js` is included in all test pages?

I just double-checked, and it turns out I actually introduced a subtle change with this PR &ndash; `localforage.js` no longer mutates `window.Promise` to polyfill; it merely contains its own promise shim. Hence the tests started failing (because they reference `Promise.all()`, `Promise.resolve()`, etc.), so I added the polyfill just for the tests. (`lie` does not require any globals.)

Technically that's a breaking change, so I'm not sure what direction y'all want to go in. My preference would be to remove the `window.Promise` mutation from `localforage.js` and also remove `nopromises.js` while we're at it, since `lie` is so small that the `nopromises` version is nearly pointless. However if you want to keep the current behavior, you can just `require('lie/polyfill')` instead of `require('lie')`, and then it's safe to remove the `es6-promise` stuff from the test HTML files. Here's an alternative PR that does just that: #553.

>  @nolanlawson OMG thank you this is pretty amazing. You rock.

:+1: You two rock @tofumatt & @thgreasi. LocalForage is amazing, and I recommend it to lots of PouchDB users who I see suffering because PouchDB is overkill for their use case. Love to see LocalForage continuing to be maintained and do well. :)

tofumatt:
So reading everything back, I think I'm happy to ditch the nopromises version. It's a small difference but a big usability/documentation/maintenance headache.

tofumatt:
Err, but, it is a breaking change, and I'd actually say we should do that as:

* a separate PR
* some kind of version bump

So I think #553 is the way to go for now, with a reversal of the contents of 5f13325584196357977333a0281c8e36f3a53198 if we want to do a single build in the future.

Issue #345, opened 2015-02-09
localforage doesn't load in node


jbuck:
Test case: `require("localforage");`

This will error out with:

```
/Users/jbuck/Documents/Github/localForage/dist/localforage.js:236
}).call(window);
        ^
ReferenceError: window is not defined
    at Object.<anonymous> (/Users/jbuck/Documents/Github/localForage/dist/localforage.js:236:9)
    at Module._compile (module.js:460:26)
    at Object.Module._extensions..js (module.js:478:10)
    at Module.load (module.js:355:32)
    at Function.Module._load (module.js:310:12)
    at Module.require (module.js:365:17)
    at require (module.js:384:17)
    at Object.<anonymous> (/Users/jbuck/Documents/Github/localForage/test.js:1:63)
    at Module._compile (module.js:460:26)
    at Object.Module._extensions..js (module.js:478:10)
```

This is because `window` does not exist in the nodejs global environment. The line to change is at https://github.com/mozilla/localForage/blob/528025b9c90ef4a66b1c0e373033cc3131ecf656/src/localforage.js#L420

I'm not sure if making localForage work in nodejs is required, but I think it'd be nice to not have it crash within node right off the bat.

fwenzel:
+1 on the "don't explode".

That said, it will be interesting to answer what it *should* do. Node does not ship with indexeddb does it? In that case I figure it would at least need to ship with one meaningful adapter out of the box that throwing data at would make sense?

psalaets:
Not sure how up-to-date the contrib guide is but see https://github.com/mozilla/localForage/blob/master/CONTRIBUTING.md#features-localforage-will-reject

fwiw localforage works with browserify

tofumatt:
Yeah, if someone can commit something that makes it work in node with little effort I guess that's fine, but right now I don't even see the use-case in node; all of the default drivers are browser-specific and localForage is aimed at browsers :smile:

Skeen:
#57 seems related to this.

Having node support makes sense if you want to utilize the same code on both client and server end.

I guess there could be a myriad of reasons to do this; among others:

* Enabling rapid prototyping and code reuse, by (temporarily) using a sub-optimal storage scheme.
    * Eventually one could port the code to use redis or another backend system.
* Testing sub-modules of a webpage with node.
    * There can be plenty of reasons for not testing within a headless browser.

Issue #57, opened 2014-02-14
Add support for Node.js


piranna:
Allow to use localForage in Node.js environment.

PR #551, opened 2016-04-19
Node.js support via browser (DOMStorage) emulation

Skeen:
This commit adds Node.js support by means of emulation DOMStorage
    (specifically localStorage).

A specific banner for node.js has been added (setting up the emulation).

The grunt 'concat' task has been expanded to generate output files for node.js,
i.e. by prepending this node.js emulation banner to the webpacked output file.

The grunt 'test' task has been expanded to include local node.js tests, these
are implemented via the subtask 'mochaTest' based upon 'grunt-mocha-test'.
- This library has been added a development dependency to package.json.

Added a ~450LOC test file to verify the functionality of localForage under node

This commit introduces a dependency on 'dom-storage', which is a tiny ~140LOC
implementation of DOMStorage inside node.

Changed the main field of package.json to point to the node output file.
Added the 'browser' field to package.json with the old main field value.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -61,6 +61,7 @@ module.exports = exports = function(grunt) {
             },
             localforage: {
                 files: {
+                    // Browser output files
                     'dist/localforage.js': [
                         'bower_components/es6-promise/promise.js',
                         'dist/localforage.nopromises.js'
@@ -68,6 +69,18 @@ module.exports = exports = function(grunt) {
                     'dist/localforage.nopromises.js': [
                         // just to add the BANNER
                         // without adding an extra grunt module
+                        'dist/localforage.nopromises.js'
+                    ],
+                    // Node output files
+                    'dist/localforage.node.js': [
+                        'src/utils/node_banner.js',
+                        'bower_components/es6-promise/promise.js',
+                        'dist/localforage.nopromises.js'
+                    ],
+                    'dist/localforage.nopromises.node.js': [
+                        // just to add the BANNER
+                        // without adding an extra grunt module
+                        'src/utils/node_banner.js',
                         'dist/localforage.nopromises.js'
                     ]
                 },
@@ -137,6 +150,15 @@ module.exports = exports = function(grunt) {
                         'http://localhost:9999/test/test.require.unbundled.html' // not officially supported since v1.3
                     ]
                 }
+            }
+        },
+        // Test node.js
+        mochaTest: {
+            test: {
+                options: {
+                    reporter: 'spec'
+                },
+                src: ['test/test.node.js']
             }
         },
         open: {
@@ -248,7 +270,8 @@ module.exports = exports = function(grunt) {
         'browserify:package_bundling_test',
         'webpack:package_bundling_test',
         'connect:test',
-        'mocha'
+        'mocha',
+        'mochaTest'
     ];
     grunt.registerTask('test:local', testTasks.slice());
 
--- a/.jshintrc
+++ b/.jshintrc
@@ -28,5 +28,14 @@
       "require": true,
       "self": true,
       "System": true
-    }
+    },
+
+    "predef": [
+        "describe",
+        "before",
+        "beforeEach",
+        "after",
+        "afterEach",
+        "it"
+    ]
 }
--- a/dist/localforage.node.js
+++ b/dist/localforage.node.js
@@ -1,0 +1,2892 @@
+/*!
+    localForage -- Offline Storage, Improved
+    Version 1.4.0
+    https://mozilla.github.io/localForage
+    (c) 2013-2015 Mozilla, Apache License 2.0
+*/
+
+/*!
+    Node.js support by emulation of browser environment (emulated DOMStorage).
+
+    DOMStorage emulation from 'dom-storage', which strives to be:
+        'An inefficient, but as W3C-compliant as possible using only pure
+            JavaScript, DOMStorage implementation.'
+*/
+// We depend on the dom-storage library
+var domStorage = require('dom-storage');
+// We utilize the './.db.json' file for persistency. 
+var localStorageEmulation = new domStorage('./.db.json', { strict: false, ws: '  ' });
+// Save the local storage implementation to the 'self' object
+self = {};
+self.localStorage = localStorageEmulation;
+
+(function() {
+var define, requireModule, require, requirejs;
+
+(function() {
+  var registry = {}, seen = {};
+
+  define = function(name, deps, callback) {
+    registry[name] = { deps: deps, callback: callback };
+  };
+
+  requirejs = require = requireModule = function(name) {
+  requirejs._eak_seen = registry;
+
+    if (seen[name]) { return seen[name]; }
+    seen[name] = {};
+
+    if (!registry[name]) {
+      throw new Error("Could not find module " + name);
+    }
+
+    var mod = registry[name],
+        deps = mod.deps,
+        callback = mod.callback,
+        reified = [],
+        exports;
+
+    for (var i=0, l=deps.length; i<l; i++) {
+      if (deps[i] === 'exports') {
+        reified.push(exports = {});
+      } else {
+        reified.push(requireModule(resolve(deps[i])));
+      }
+    }
+
+    var value = callback.apply(this, reified);
+    return seen[name] = exports || value;
+
+    function resolve(child) {
+      if (child.charAt(0) !== '.') { return child; }
+      var parts = child.split("/");
+      var parentBase = name.split("/").slice(0, -1);
+
+      for (var i=0, l=parts.length; i<l; i++) {
+        var part = parts[i];
+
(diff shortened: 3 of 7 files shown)


Skeen:
Hello,

I am aware that this pull request will be rejected, according to the;
    'Features localForage will reject' --> node.js support.

Yet I'm proposing this pull request for others to find, when they require node
support in the future. I created this since I needed node support for two 
reasons, namely;

* Running the same code on both client and server.
    * This enables rapid prototyping by maxizing code reusability.
    * It helps out new programmers, as they only need to learn one new library.
    * As the library has a simple key-value interface, it's easy to eventually
        port code to redis or another more optimal backend key-value storage.    

* Testing browser code locally (outside of the headless browser configuration).
    * One case where this makes sense, is by the above, i.e. testing local code,
        which utilizes this library outside of the browser context.
    * Another case can be when resources or time is limited, testing directly
        in node is usually faster than within a headless browser environment.

Nevertheless; this pull request addresses: #345 and #57.

Skeen:
While writing my unit tests; I found two pieces of unexpected behaviors.

I have not tested the browser behavior yet. As such I cannot tell if the they 
are reproducable, or they stem from bugs in the emulation layer.

I just want to (de)confirm my understanding of the API.

The two cases are:

* removeItem

```js
it('should not allow removing invalid keys', function(done) {
    localforage.removeItem('key', function(err)
    {
        ...
    });
});
```
Being run on an empty storage, I would expect an error (we're trying to remove,
a non existing key).

* keys

```js
it('should start out empty (value check)', function(done) {
    localforage.keys(function(err, keys)
    {
        ...
    });
});
```
Being run on an empty storage, I would expect keys to be '[]', but this is not
the case.

thgreasi:
That's great and simple! If this doesn't get merged, then we should make a
custom driver for it. I can start experimenting with such implementation if
you are fine with it. We could probably create a github organization like
`localforage-extensions` if @tofumatt is ok with such naming.

> --
*Thodoris Greasidis*
Computer, Networking &
Communications Engineer

tofumatt:
Wow, this is actually quite cool. I agree, we shouldn't reject this on principle if it's not too intrusive to the code. I'll have a review soon.

- tofumatt

Skeen:
@thgreasi: A custom driver would be fine with me; all I really wanted was node.js support and hence I went with the easiest and quickest implementation that would fulfill my requirements.

- The argument being that utilizing a client-side library for server-side is somewhat 'wrong', and hence you should not expect superb performance or metrics, only correctness.

@tofumatt: When during your review please keep a few things in mind, which may be deal-breakers:

The `package.json` 'browser field' has been added, and the 'main field' has been changed; [package.json:55+56](https://github.com/mozilla/localForage/pull/551/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R55)

- This affects the ordinary functionality of the library, i.e. it COULD break stuff. Thorough testing would be required; also the 'browser field' is not even in the real package.json standard, but rather posted [here](https://github.com/defunctzombie/package-browser-field-spec).

     - We specifically need to check, whether the bundlers we want to support actually respect it. According to [this](https://gist.github.com/defunctzombie/4339901#gistcomment-1582456) `webpack` may need a work around, while `jspm` and `browserify` seem to respect it as is.

The code should be modified to pass the jshinter, such that the actual testing errors mentioned [here](https://github.com/mozilla/localForage/pull/551#issuecomment-211659495), actually show up on travis.

- I will look into this first thing, when I get into the office in a few hours. We need to discuss these diverging test cases, but before looking further into these; I'll test the correspondence with the browser implementation.

We need unit tests for 'iterate' and the entire promises based Data API.

Skeen:
Travis now builds and fails to two mentioned test cases.

I have tested these two test cases on in the browser environment (via. [js-fiddle](http://jsfiddle.net/));

* removeItem

The browser does not throw an error when removing a non-existent element, so the node.js implementation seems in sync with the browser implementation.
-- I'm curious however as to why this does not happen?

* keys

The browser returns an empty array '[]', when there are no keys stored. This is the behavior I expected, and hence I conclude that the failure is in the node.js implementation. I'll look into this.

Skeen:
`keys` did in fact return an empty array on node.js as well, my method of checking was improper. As such the node.js implementation seems to be completely in-sync with the browser one, in regards to the provided unit tests.
-- We still miss unit tests for `iterate` and the entire promise based data api.

nolanlawson:
Agree this is a nice feature to have, ideally without needing a custom adapter. It'd be good to double check that this is the best possible implementation, though, since in Node we have more options than just a Local Storage shim.

Gruntfile.js
```diff
@@ -69,6 +70,18 @@ module.exports = exports = function(grunt) {
...
+                    'dist/localforage.node.js': [
+                        'src/utils/node_banner.js',
+                        'bower_components/es6-promise/promise.js',
+                        'dist/localforage.nopromises.js'
+                    ],
+                    'dist/localforage.nopromises.node.js': [
+                        // just to add the BANNER
+                        // without adding an extra grunt module
+                        'src/utils/node_banner.js',
+                        'dist/localforage.nopromises.js'
```
nolanlawson: This is a bit of an unsatisfying pattern, because `dist/ `files (i.e. bundle files) are not typically used for Node dependencies.

Ideally people should be able to do:

```js
var localforage = require('localforage');
```

And this should "just work" in either the browser (via browserify/webpack/etc.) or Node. Typically the way to achieve that is with `"browser"` flags in `package.json`: e.g. [here's how PouchDB does it](https://github.com/pouchdb/pouchdb/blob/985f548f2cbd818ed0158ad8f97da31c79485654/package.json#L120-L128).

Also I'd say a `nopromises` version for Node is unnecessary, because large bundle sizes are less of an issue on the server side.

nolanlawson:
I just took a look at [dom-storage](https://github.com/node-browser-compat/dom-storage), and based on the implementation, I would strongly advise against using it in LocalForage. It advertises itself as mostly designed for unit testing, and that shows in [the source](https://github.com/node-browser-compat/dom-storage/blob/5670628570c7d744bd78735788f7fd8c7e3870bc/lib/index.js).

Notice that it works by relying on `fs.readFileSync`/`fs.writeFileSync`/etc., which is slow and synchronously blocking. For LocalForage to have a good reputation as a storage engine that works well in all environments, this would be a poor choice for a default.

As an alternative suggestion, how about:

1. Writing a custom adapter that uses [node-websql](https://github.com/nolanlawson/node-websql) via the WebSQL adapter, or that uses [level](https://www.npmjs.com/package/level) via a custom implementation, then
2. opening a PR to use that adapter by default in Node and then use the other three adapters in the browser.

Another alternative is to just rewrite the existing WebSQL adapter to use `node-websql` (which does the appropriate `"browser"` switching automatically), in which case the only change to localforage core would be to set the list of adapters to `[idb, websql, localstorage]` in the browser and `[websql]` in Node.

Skeen:
I guess I have to agree with @nolanlawson, due to the quality of [dom-storage](https://github.com/node-browser-compat/dom-storage), maybe this pull request should be rejected.

I don't mind this; I expected this pull request to be rejected on principle.
I just wanted to publish my proof-of-concept for others to find.

@nolanlawson suggestions for alternative implementations sound excellent to me, but quite frankly my patch fulfills my own requirements (proof-of-concept prototyping), and I'm afraid I wont be implementing he's suggestions.

Gruntfile.js
nolanlawson: Actually it just occurred to me that we can solve this problem by simply changing the `"main"` in the `package.json` to point to the Node distributable version.

nolanlawson:
Sorry, I wasn't trying to make you feel unwelcome. :( I think this PR demonstrates a desire to have localforage run in Node, and we should definitely move in that direction.

Skeen:
Oh, I do not feel unwelcome at all! ^.^

I'm just stating that my desire to have localforage run on Node was purely for proof-of-concept projects, not for production quality projects, which shows by the fact that I decided to utilize a somewhat random localStorage shim rather than implement a better solution.

Hence, I simply agree with you, that while dom-storage might be fine for proof-of-concept kinda deals, it is indeed not a production quality library.

A production quality implementation would be nice to have, but it was simply not worth the effort for my use case :).

Gruntfile.js
Skeen: [I did that](https://github.com/mozilla/localForage/pull/551/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R55), and I agree that the `nopromises` version may be a bit silly for node.

Skeen: To replace the dist files (i.e. for getting the header in there), what would be a better approach;

* Making an `index.js` file for nodejs, which includes the node.js header and the rest of the library?

pwFoo:
I don't know how it works, but there is "A drop-in substitute for the browser native localStorage API that runs on node.js"
https://github.com/lmaccherone/node-localstorage

PR #553, opened 2016-04-23
re-enable window.Promise polyfilling

nolanlawson:
Alternative version of #538 that re-enables
the automatic `window.Promise` polyfilling, as
currently done in localforage.


nolanlawson:
More explanation for this PR: https://github.com/mozilla/localForage/pull/538#issuecomment-213813380

thgreasi:
Great instant alternative PR!
Is there a reason that the ternary check of whether Promises are available
or not got removed? I'm worried about the nopromise build and how requiring
an excluded module (lie/polyfill) would work (even on browsers that
Promises are available).

I saw that lie also polyfills other APIs, like Mutation Observers. I'm
impressed about how small it is, compared with the es6-promises that we
used until now!

PS: Perhaps we should have added a test cases for the nopromises build when
it was first introduced, where we would polyfill Promises manually before
including localforage.nopromises.js .

> --
*Thodoris Greasidis*
Computer, Networking &
Communications Engineer

tofumatt:
Sorry, I merged some changes to the `Gruntfile.js` in #552 and it made this not merge cleanly.

nolanlawson:
@tofumatt - Re-rebased this and the other PR, no worries.

@thgreasi:

> Is there a reason that the ternary check of whether Promises are available or not got removed?

Yep, if we're using `require('lie/polyfill')` rather than `require('lie')`, then the global `Promise` object gets mutated automatically and there's no reason for a ternary check.

>  I'm worried about the nopromise build and how requiring an excluded module (lie/polyfill) would work

In this case, `require('lie/polyfill')` becomes a no-op, and then `export Promise` exports the global `Promise` object. So it will fail in browsers without Promises, or cases where the user has not provided their own global polyfill, which matches the current behavior.

PR #555, opened 2016-05-02
Create memory.js for fallback

rubenstolk:
- [x] This helps for running localforage without issues in e.g. private browsing modes
- [x] It simply stores data in memory, which will of course be lost after refreshing the page


tofumatt:
Very cool, though I wonder about including this. Users should expect that localForage saves their data permanently, and if it can't then it throws an error/rejects a Promise and they can deal with that. Using the memory backend, we won't alert the user that there's a problem saving data permanently (even if that problem is private browsing mode). Alerting the developer that we couldn't save data, and then allowing them to bubble that info up to the user could be useful.

On the other hand, it makes using localForage way more annoying--your patch makes the experience of just saying "Private Browsing means no permanent data" work as intended for all. And developers can always do their own checks for Private Browsing mode if that matters.

Side note: looks like adding the driver causes these ["no driver available" tests to fail](https://travis-ci.org/mozilla/localForage/builds/127197507#L594), which makes sense. We'll also want to add tests for the in-memory driver.

I've come around to this idea I think.

Gruntfile.js
```diff
@@ -25,6 +25,7 @@ module.exports = exports = function(grunt) {
             'src/utils/serializer': 'localforageSerializer',
             'src/drivers/indexeddb': 'asyncStorage',
             'src/drivers/localstorage': 'localStorageWrapper',
+            'src/drivers/memory': 'memoryStorageWrapper',
```
tofumatt: I would just call it `memoryStore` instead of `memoryStorageWrapper`.

src/drivers/memory.js
```diff
@@ -0,0 +1,247 @@
...
+            resolve(value);
+        });
+    });
+
+    executeCallback(promise, callback);
+    return promise;
+}
+
+var memoryStorageWrapper = {
+    _driver: 'memoryStorageWrapper',
```
tofumatt: I'd call it `memory` instead 😄

rubenstolk:
@tofumatt I just made this in 30 minutes because I was in a hurry for one of our clients. Leaving for holiday now for a week. If you feel like including it the way you want, I'm open for it, otherwise I'd love to update the PR later with tests et cetera, but basically just wanted to pass the idea with you first, hence the early-stage PR :)

tofumatt:
Cool! Yeah, let’s plan to iterate on this a bit when you come back. Go enjoy your holiday 😉👍🏻

rubenstolk:
Perhaps we could include it but have an explicit "fallback_to_memory" option so the developer can decide... Yep, I'll be back on it!

thgreasi:
I actually had in mind to create a custom driver for this, but I always postponed it. :+1: for opening this PR and discussion.
The main benefit of a custom driver is that we can avoid creating an extra `dist` file (or increase the size of an existing one) and have the driver optionally included only by those who need it.
As @tofumatt mentioned, I like the freedom/flexibility that this offers to the developer:
```js
var localforageReady = localforage.ready().catch(function(err) {
  if (err.message == 'No available storage method found.') {
    localforage.defineDriver(memoryStorage);
    return localforage.setDriver(memoryStorage._driver).then(function() {
      return localforage.ready();
    }
  }
});
```

PS: After some digging around, I found a primitive (pre ES6) implementation like this in the [`dummyStorageDriver` used in the test cases](https://github.com/mozilla/localForage/blob/master/test/dummyStorageDriver.js)!

nolanlawson:
Quick suggestion: instead of writing a whole new adapter, we could also just rewrite the localstorage adapter to use @gr2m 's  [humble-localstorage](https://github.com/gr2m/humble-localstorage).

It automatically switches to in-memory mode when localstorage is unsupported, e.g. in private browsing mode. It's also very smart about detecting private browsing mode, e.g. in weird environments like Safari where localstorage basically acts like `/dev/null` and you have to test it to see if it's actually storing anything.

nolanlawson:
As a side note, would it be possible to distribute this as a third-party adapter? It seems like exactly the kind of thing that would be more useful as a separate plugin than as something that increases the core size of the library, while adding features that not all LocalForage users would necessarily desire.

tofumatt:
Just a thought: when fallback to memory.js happens, we could include a console warning saying the dev is likely in private browsing mode or drivers aren't available.

Gruntfile.js
fbender: Or better yet make it something with "volatile memory" so it is absolutely clear this is not persistent!

thgreasi:
> As a side note, would it be possible to distribute this as a third-party adapter?

In the mean time, as a way expand the discussion (and possibly as a boilerplate), I just restructured and pushed the dummyStorageDriver as a custom [memoryStorageDriver](https://github.com/localForage-extensions/localForage-memoryStorageDriver).

rubenstolk:
I guess I would agree with having this as a separate module... Shall we close this PR in favor of @thgreasi's solution?

rubenstolk:
So, unfortunately as rightfully indicated by @thgreasi in https://github.com/localForage-extensions/localForage-memoryStorageDriver/issues/2#issuecomment-219037643, the original problem of this PR can't be solved with a custom driver.

How about including the memoryDriver in `localForage` but as an optional spec which distributes into `localstorage.with-memory-storage.js`?

thgreasi:
How about exporting the [`DefaultDriverOrder` array](https://github.com/mozilla/localForage/blob/44194a81b92664d651fb8855aae86c7af3ce7412/src/localforage.js#L21-L25) so that plugins can manipulate it?
Obviously, there are no guaranties whether we completed the manipulation before the default `localforage` instance settled to a driver, but since localforage settles on a driver lazily, we can be almost sure about it, if the manipulation is done right after importing the scripts.

bsr203:
just exploring to switch from https://github.com/google/lovefield to this. I too like to have memory driver mainly for running my tests in `jsdom` which doesn't support offline storage mechanisms. Will explore @thgreasi 's sol. thanks.

jonstorer:
@tofumatt first off, sorry for the down vote, github is getting pretty social.

What I mean to say was that a console error won't really suffice. What about an even to bind to? Something that would allow to dev to bind a message to the user?

Something like "we noticed you're in private browsing mode. That's cool, but things will be slower and nothing will be saved when you refresh". 

Point being, the dev can decide how to prompt the user... instead of just knowing it's happening.

jonstorer:
I'm new to the discussion, so take my input with a grain of salt.

I think that a memory strategy include within the core of the library is the correct solution. The point of this library is to ensure that some variable of local storage is available regardless of browser or version. 

I also think that including some strategy to bind to this error to allow the dev to handle the interaction with the use gracefully solves most of the concerns above.

@tofumatt to your points about "not really saving data". You're correct. However, an assumption I'm making is that the library's goal is to make browser storage work regardless of anything.

AshleyScirra:
I mentioned this PR here: https://news.ycombinator.com/item?id=13422108

I am also very much in favour of in-memory fallback as a core feature. It's super annoying how Firefox throws exceptions if you touch storage in private mode. It tends to break pages. AFAIK Chrome Incognito allows storage to work but just keeps it in RAM and throws it away on the next navigation away, so this feature is simply replicating what a major browser does and making it consistent across more finicky browsers.

AshleyScirra:
Incidentally just filed issue #647 which I think proves the point that this ought to be built-in.

gabrielmaldi:
> So, unfortunately as rightfully indicated by @thgreasi in localForage/localForage-memoryStorageDriver#2 (comment), the original problem of this PR can't be solved with a custom driver.
> 
> How about including the memoryDriver in localForage but as an optional spec which distributes into localstorage.with-memory-storage.js?

This would be solved by #692.

thgreasi:
@rubennorte thanks a lot for the initial PR on this.
Rebased & merged with 40c70210a54645c33919a283b2a1138f40b7a001 on the v2.0-dev branch.
Too bad GH didn't detect that as a merge, probably b/c of changing the base branch.

See: https://github.com/localForage/localForage/commit/1e8828371136e4779521c91f2b2c30bf6381f10f
See: https://github.com/localForage/localForage/commits/v2.0-dev?after=40c70210a54645c33919a283b2a1138f40b7a001+69

Issue #445, opened 2015-09-05
Error: Refused to connect to 'blob:http


siddo420:
I am getting the following error message in Chrome

```
Refused to connect to 'blob:http%3A//192.168.1.9%3A8080/3547f91f-dd73-48a4-9bf1-233c183c22ed' because it violates the following Content Security Policy directive: "connect-src 'self' wss://192.168.1.13 https://192.168.1.13".

_createBlob.type @ localforage.js:1036
_blobAjax @ localforage.js:1034
txn.oncomplete.getBlobReq.onsuccess @ localforage.js:1087
```

JS and HTML are served from http://192.168.1.9:8080

tofumatt:
Can you include the JS you are running that's causing the error and also your server's CSP rules? This looks like a CSP directive violation, perhaps you're trying to store or retrieve data you aren't allowed to interact with in JS.




I can see the blob was retrieved successfully as the error happens inside the onsuccess handler.



- tofumatt

coolaj86:
I'm having the same issue with this snippet:

```html
<!DOCTYPE html>
<html>
  <head>
    <title>LocalForage Test</title>
    <meta http-equiv="Content-Security-Policy" content="default-src 'self'; connect-src 'self' http://conf.utahjs.com/; font-src 'self' https://fonts.googleapis.com/css/ https://fonts.gstatic.com/s/raleway/; style-src 'self' https://fonts.googleapis.com/css; child-src 'none'; object-src 'none'; img-src 'self' https://avatars.githubusercontent.com/ https://avatars.io/email/ https://avatars.io/twitter/ https://www.gravatar.com/avatar/">
    <link rel="stylesheet" href="bower_components/bootswatch-dist/css/bootstrap.css" />
  </head>
  <body>
    <script src="bower_components/localforage/dist/localforage.js"></script>
    <script src="lf-test.js"></script>
  </body>
</html>
```

```js
(function () {
  'use strict';

  var store = window.localforage.createInstance({
    name: "conf"
  });

  store.getItem('schedule').then(function (err, sched) {
    if (err) {
      console.error(err);
    }
    if (sched) {
      console.log(sched);
    }
    console.log('blah');
  });

  store.setItem('schedule', {});
}());
```

![screen shot 2015-09-12 at 4 42 53 pm](https://cloud.githubusercontent.com/assets/122831/9834192/7051bd14-596d-11e5-8f5b-aedbdbe2ec1f.png)

coolaj86:
[...]

PR #558, opened 2016-05-03
refactor(localforage): simpler blob support check

nolanlawson:
Rather than doing an ajax request for a blob, which shows up in the user's console and tends to cause confusion (or may even cause security exceptions in some environments; I suspect #445 is related), just do a user-agent check to determine if we're running in Chrome 37-42, which are the versions that allow you to store a blob but actually have bugs that end up breaking later down the line.

This is the same check we now use in PouchDB, both for the code simplification and to avoid the ajax request. We've been using it for about a year now with no complaints from users, so the user agent check seems to be foolproof.

You can also test this in Saucelabs (as we do) by testing against Android 5.1. Android Lollipop ships with Chrome 37 by default, and Sauce does not update the Chromium WebView in their images, meaning you can reproduce the error fairly easily that way.

### DIFF
--- a/src/drivers/indexeddb.js
+++ b/src/drivers/indexeddb.js
@@ -25,75 +25,42 @@ function _binStringToArrayBuffer(bin) {
     return buf;
 }
 
-// Fetch a blob using ajax. This reveals bugs in Chrome < 43.
-// For details on all this junk:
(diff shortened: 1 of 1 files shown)


tofumatt:
That's totally nicer, even if it's a UA check, thanks 👍

PR #563, opened 2016-05-14
Check import lie

### LINKED ISSUE #562: v1.4.1 - Error: Cannot find module 'lie/polyfill'
I believe this is a bug in v1.4.1 (nopromises), specifically at line 105.
`_dereq_('lie/polyfill');`

I suspect this line expects the client to be running CommonJS. On web browsers, this fails and localforage fails to load.

Once line 105 is removed, localforage appears to work.

thgreasi:
Fixes #562


tofumatt:
wfm

nolanlawson:
wtf, the "no promises" build is specifically supposed to skip `lie`. that's the whole point.

I will look into this. it's probably the distinction between `lie/polyfill` and `lie` that messed things up, my bad

src/utils/promise.js
```diff
@@ -1,4 +1,6 @@
 // This is CommonJS because lie is an external dependency, so Rollup
 // can just ignore it.
-require('lie/polyfill');
+if (typeof Promise === 'undefined' && typeof require !== 'undefined') {
```
nolanlawson: the reason this won't fix the underlying issue is that it will still fail on non-Promise browsers, right?

thgreasi:
> it will still fail on non-Promise browsers, right?

That's the point of the nopromises build, that it fails on older browsers, except if the an appropriate polyfill is provided by the user.
You are probably right about the `require` check part, added it just in case.
Before this change, even the index.html found in the examples folder failed with the nopromises build on my Firefox dev edition, since the derequired require() call failed. So, I ended up adding a check before the invocation.

thgreasi:
To be clear: a check wether Promises are already available.

thgreasi:
Nevermind, just saw your PR :)

PR #586, opened 2016-07-02
Use Object.prototype.toString for type checks

duzun:
Consider the following code using 'LOCALSTORAGE' driver:

```js
localforage.setItem('test', '[object Blob]')
.then(() => {
    localforage.getItem('test');
});
```

This is 100% legitim use case. But it fails because the way serialize() checks value's type.

Basically instead of `value.toString()` it should be `Object.prototype.toString.call(value)`, which yields `"[object String]"` instead of `"[object Blob]"` for the above example.

Another potential issue is `value instanceof Blob`. 
I see two scenarios where it would get wrong:

1. `Blob` type doesn't exist (blob not supported);
2. `value` comes from a different window context (eg. an iframe).


tofumatt:
Hi there!

This looks good but sadly it has merge conflicts with `master`. Could you rebase your changes against master and push again? Thanks!

duzun:
No problem!

duzun:
Done!

tofumatt:
You rock, thanks very much!

duzun:
You are welcome!

PR #587, opened 2016-07-08
Fix bug in indexeddb detection code

tln:
I ran into an issue when storing Blobs with Chrome 51 where part of the detection expected a transaction but was getting an IDBDatabase. I haven't created extra test cases to catch this, I should be able to do that in the next couple of weeks if needed.


src/drivers/indexeddb.js
```diff
@@ -70,7 +70,8 @@ function _checkBlobSupport(idb) {
     if (typeof supportsBlobs === 'boolean') {
         return Promise.resolve(supportsBlobs);
     }
-    return _checkBlobSupportWithoutCaching(idb).then(function(value) {
+    var txn = idb.transaction(DETECT_BLOB_SUPPORT_STORE, 'readwrite');
```
tofumatt: Can you expand this variable name? I assume it means `transaction`, but I like verbose variable names 😄

tofumatt:
Looks good; it would be great if you could add a test for this, and you'll need to rebase again master. But thanks for the patch, I think we can merge it once you add a test, thanks!

martinleopold:
Ran into that same issue just now and came up with pretty much the same exact patch.

This bug results in blob support not being detected correctly, thus blobs are serialized and this ultimately triggers serialization size limits (~127MB. I'm also on Chrome 51).

>  DOMException: The serialized value is too large (size=258433118 bytes, max=133169152 bytes).

I'd be happy to try and provide a test if you point me in the right direction of where/how to best add it, @tofumatt.

nolanlawson:
Wow this is an awful bug that slipped through; I'm super sorry about that.

@tofumatt I don't think this needs an additional test case, because it looks like the bug is causing the indexeddb adapter to do the base64 serialization everywhere, which of course will pass the tests. Testing it would require reaching into IndexedDB and confirming blob vs base64, which will vary from browser to browser.

dac09:
I'm seeing a similar issue on `Chrome 52.0.2743.116 (Official Build) (64-bit)`

It seems to switch between indexDB and webSQL. So when I create a new instance, and try to remove keys, it fails silently.

tofumatt:
I am extremely sorry this took so long to get to but I'm gonna fix the conflict and merge this in. Thank you.

thgreasi:
Ooops I think that I resolved the issue by doing something similar in #643 . I'm really sorry about not merging this one instead :(

tofumatt:
Ah okay, I was gonna say I feel like this was fixed and was surprised it wasn't merged. I'll close this then, no worries.

PR #592, opened 2016-07-15
Added osx support to travis ci

amilajack:


tofumatt:
Not sure what this gets us right now, what's the reason for this?

In the future when we're running karma tests (#461) I'm sure we'll want this, but right now it will only make tests run slower with no benefit, right?

amilajack:
Sure thing. I'll close PR

PR #595, opened 2016-07-24
fix(utils/idb): fix SecurityError on FF

### LINKED ISSUE #501: SecurityError in Firefox when third party cookies are disabled
localForage is used on a web page that is iframe embedded on a second web page. This causes driverSupport to throw SecurityError (SecurityError: The operation is insecure.) Note: Firefox is used and third party cookies have been disabled.

The error occurs when driverSupport is accessing self.indexedDB. A possible solution could be to add try/catch when accessing self.indexedDB.

```javascript
// OLD CODE
//var indexedDB = indexedDB || self.indexedDB || self.webkitIndexedDB || self.mozIndexedDB || self.OIndexedDB || self.msIndexedDB;

// NEW CODE
var indexedDB = indexedDB;
if (!indexedDB) {
    try {
        indexedDB = self.indexedDB;
    } catch (e) {
        indexedDB = undefined;
    }
}
if (!indexedDB) {
    indexedDB = self.webkitIndexedDB || self.mozIndexedDB || self.OIndexedDB || self.msIndexedDB;
}
```

thgreasi:
Resolves #501.
Based on the suggestion of @daviddeutsch.

### DIFF
--- a/src/utils/idb.js
+++ b/src/utils/idb.js
@@ -1,20 +1,22 @@
 function getIDB() {
     /* global indexedDB,webkitIndexedDB,mozIndexedDB,OIndexedDB,msIndexedDB */
-    if (typeof indexedDB !== 'undefined') {
-        return indexedDB;
-    }
-    if (typeof webkitIndexedDB !== 'undefined') {
-        return webkitIndexedDB;
-    }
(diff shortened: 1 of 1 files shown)


tofumatt:
Looks good to me!

Issue #380, opened 2015-05-25
ability to store Blobs (etc.) nested within other objects


sheppard:
I'm working on integrating `localForage` into [wq.app](https://wq.io/wq.app) and overall have found it quite useful.  I'm using the blob storage capability to enable users to take photos with their phones offline, and store them in an "outbox" until they are able to connect.  For this to work, I'd like the ability to store nested records of the form:

```javascript
{
    "id": 123,
    "date": "2015-05-24",
    "photo": {
        "type": "image/jpeg",
        "name": "IMAG0001.jpg",
        "body": Blob(type="image/jpeg", size=20104)
    }
}
```
This works fine on the `IndexedDB` driver, but for the serializing drivers the full `Blob` is not saved unless it is the top level object.

I was able to work around this in [wq/store.js](https://github.com/wq/wq.app/blob/master/js/wq/store.js#L338-L528) by making a `setItem()` wrapper that recursively extracts all blobs from an object into separate keys (and then puts them back in place during `getItem()` so the user never knows the difference).  Would a variation on this algorithm be appropriate for inclusion in the `localForage` serializer?

tofumatt:
I would say this code would be welcome; it allows users to store blobs are part of any object easily. My only concern is performance--doesn't this mean we need to deep scan every Object to check for blobs on set and get even in IndexedDB (which we didn't before).




If the performance hit isn't huge it's worth it though.



- tofumatt

sheppard:
Ok, I'll think a bit more about how this would work and see if I can put something together.  I see two potential approaches:
 1. Handle in the serializer (convert to/from base64 in-place)
 2. Handle outside of the serializer (offload to separate keys in whatever driver)

What I've [implemented in wq/store.js](https://github.com/wq/wq.app/blob/60f17c4add0654d1f38a3eb4d2b45911b0af01b4/js/wq/store.js#L379-L392) is more like the second option.  It's automatically [disabled for IndexedDB](https://github.com/wq/wq.app/blob/60f17c4add0654d1f38a3eb4d2b45911b0af01b4/js/wq/store.js#L92-L97) unless the user explicitly requests blob extraction.

[...]

PR #603, opened 2016-09-14
search for and serialize blobs nested within other objects

sheppard:
Here's an implementation of #380.  I went with the first option discussed there (handle in serializer) since it makes the implementation much simpler (especially with the refactored serializer module).  The second option discussed in #380 would have allowed customizing where blobs were stored, but I've ended up not actually using that ability in the applications I've built.  It's also a bit of a mess to deal with swapping out references and trying to keep everything synced in an atomic way.  Doing everything in the serializer instead seems to make the most sense.

As discussed in #380, this feature requires a deep scan to find any nested blobs.  It doesn't feel like a big performance hit but I can imagine some users might not want it.  I added a flag (`localforageSerializer.options.recursive`) for the feature and set it to `true` by default.  I'm not quite sure what the default should be, or where to provide a public API for overriding it - any ideas?


tofumatt:
I’ve been extra busy the past week and will be probably until the end of
the month, but know I do plan to take a look at this, cheers!

Narcwis:
I am working on a project which saves an object with two properties (Blob and Date). I have tried your solution and indeed this does save blobs and retrieves them but it empties the date property as per screenshot. 
image

Previous behaviour is documented in [this comment](https://github.com/localForage/localForage/issues/604#issuecomment-276800409). This issue is present on Safari for macOS only. Testing of above was done on both Safari 10.0.3 and 10.2.

RaulTsc:
@tofumatt any news about this? :)

PR #606, opened 2016-10-01
docs(README): add ES6 Typescript import style notice

### LINKED ISSUE #605: Typescript section seems to be outdated?!
The typescript section in the README file seems to be outdated. I don't have any issues importing localforage using the ES6 modules syntax. The latest commit from @thgreasi might be the reason for that. 

https://github.com/DefinitelyTyped/DefinitelyTyped/commit/eaeea6e83fc58f97a789813de62e8a387dd5fe97

I tested it with TS 1.8 as well as 2.0.

thgreasi:
Resolves #605 while keeping the `import = require()` style for users of older versions of Typescript.

### DIFF
--- a/README.md
+++ b/README.md
@@ -171,16 +171,17 @@ installed (eg. `npm install --save-dev babel-plugin-system-import-transformer`).
 
 ## TypeScript
 
-To import localForage in TypeScript:
+To import localForage in TypeScript use:
+
+```javascript
+import * as localForage from "localforage";
+```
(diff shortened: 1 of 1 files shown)


thgreasi:
@iamolivinius does this looks fine to you as well?

iamolivinius:
Maybe add another line like 

import {setItem, getItem} from 'localforage'

Besides that... LGTM :)

Issue #600, opened 2016-08-28
Add typescript definitions to the repo


whitecolor:
Why not add type definition directly in the repo and package for God's sake and great goodness?

https://github.com/DefinitelyTyped/DefinitelyTyped/pull/10859

thgreasi:
I'm totally fine with such an idea. @tofumatt what's your call? Such a thing would allow us to avoid downloading the huge DefinitelyTyped repo every time we need to make a change. Alternatively we could create a separate localforage.d.ts repo.

PS: atm I'm thinking of incorporating the `export =` syntax in the typings, just like how pouchdb did. CC: @nolanlawson

tofumatt:
Sounds fine to me, though I could certainly comment more if I saw a PR.

But I assume it's just a file that goes in the root of the project? I'm just not a typescript user, so if it has to be maintained someone would need to document how to do that 😄

whitecolor:
@tofumatt yes its just a file and "types" filed in `package.json`

@thgreasi 
> PS: atm I'm thinking of incorporating the export = syntax in the typings, just like how pouchdb did.

will you able to accomplish this with PR?

You also have this nice plugins (setItems, getItems, removeItems) https://github.com/localForage
I'm not sure how they should play and could they extend with localforage typings.

tofumatt:
Cool then, works for me!

thgreasi:
@whitecolor What troubles me is how the various localforage-extensions will be able to reference the `localforage.d.ts` (eg a custom driver should be able to reference the `LocalForageDriver` type), since the [`Typings for NPM Packages` docs](https://www.typescriptlang.org/docs/handbook/typings-for-npm-packages.html#your-definition-files-should) state that:
> Your definition files should ... not contain triple-slash references.

In the case of DefinitelyTyped, that was not a problem since all the definition folders are defined next to each other. Any ideas?

whitecolor:
@thgreasi 
> Your definition files should ... not contain triple-slash references.
Triple slashes seems to me obsolete 

I asked the question here:
https://github.com/Microsoft/TypeScript/issues/10635

> What troubles me is how the various localforage-extensions will be able to reference the localforage.d.ts

I my view it should be done just:
```js
import {LocalForageDriver, localforage} from 'localforage'
```
It is achieved by just exporting interfaces/types

thgreasi:
Opened PR #611 for this.

Elephant-Vessel:
[...]

PR #608, opened 2016-10-07
docs(README): update Typescript notice for v2.0

thgreasi:


thgreasi:
@iamolivinius does this looks fine to you as well?

iamolivinius:
Funny you mention it. I was tinkering with that syntax, too.

Unfortunately the compiler (typescript@2.0.3) throws an error for

```typescript 
import localForage from "localforage";
```

```sh
main.ts(1,8): error TS1192: Module '"localforage"' has no default export.
```

This is totally reasonable when you read the type definitions here: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/localForage/localForage.d.ts#L114 
and: https://github.com/Microsoft/TypeScript-Handbook/blob/master/pages/Modules.md#export--and-import--require

I think an update of the type definitions is needed as well to get everything in line.

thgreasi:
I got it working yesterday on a [test Ionic 2 RC0 project](https://github.com/thgreasi/localForage-cordovaSQLiteDriver-TestIonic2App). I was struggling with it the whole week, but never made it work with the `import * as localforage from "localforage"` syntax (which worked fine in Ionic2 beta & probable pre v2.0). The syntax in this PR is also mentioned in a few ["working" comments in the Ionic forums](https://forum.ionicframework.com/t/where-is-storage-and-localstorage-in-rc0/64886/16).

Can you test whether it woks for you?
As of the typings, we can't make define a default export since we are a UMD module (not an es6 module). We tried that some time ago but didn't work as expected, so we now use the `export =` syntax which better matcher non-es6 modules, (following the lead of the PouchDB guys) . The only extra thing to do, is probably use the [`export as namespace`](https://github.com/Microsoft/TypeScript/wiki/What's-new-in-TypeScript#support-for-umd-module-definitions) syntax that typescript 2 added.

thgreasi:
Can you check whether this works for you on ionic 2 RC0?
In my case it seems to work:
* no errors on sublime
* no compiler errors
* works with ionic serve
* works when emulating on android

thgreasi:
Maybe what is mentioned in #609 is the reason that `import localforage from "localforage"` works on ionic 2 rc0, even though earlier versions worked with the `import * as localforage`.

imcotton:
I think `import * as localforage` only gives you `localforage = {default: 'instance of LocalForage'}` at runtime, but IDE would provides you code completion to variable `localforage`

imcotton:
Continues in [#609](https://github.com/localForage/localForage/pull/609#issuecomment-252424753)

> I would prefer targeting the default configuration of each Typescript version.
> In case a v1.8 user changes his allowSyntheticDefaultImports option, it might cause other used libraries to stop building/working.

`allowSyntheticDefaultImports` is flag to let TS code completion treat it as a default ES6 import, it shouldn't changes the output result.

>ES6 Style imports are not working.

>When you want to use ES6 style imports but the typings do not yet use ES6 style exports, then set the TypeScript compiler option allowSyntheticDefaultImports to true.

ref:
- https://code.visualstudio.com/docs/languages/javascript#_troubleshooting-intellisense
- https://code.visualstudio.com/docs/runtimes/nodejs#_adding-a-jsconfigjson-configuration-file

thgreasi:
Just updated this PR after @imcotton's awesome hints in #609.
Waiting for your feedback.

imcotton:
Nice, maybe also have `@types/localforage` installation mentioned?

iamolivinius:
Typescript compiles this

```ts
import lf from 'localforage'
lf.getItem('1234')
```

to that, right?

```js
"use strict";
var localforage_1 = require('localforage');
localforage_1.default.getItem('1234');
```

imcotton:
@iamolivinius 
That's true for TS if `--module` targeting to `commonjs`/ `amd` / `system` / `umd`
By targeting to `es6` / `es2015` will leave code untouched and let **Rollup** or **SystemJS** to do the same thing.

iamolivinius:
The problem I see is that the files under dist/ do module.exports only. I think export.default is missing for some reason.

imcotton:
Not necessarily.
`module.exports = localforage_1 ` in CJS is equivalent to `export default localforage_1` in ESM.

thgreasi:
@imcotton there is also #600 that suggests that we should add the typings as part of this repo, so I though about not adding `@types` notice in this PR.
To all, thanks for your elaboration on this! Never seen such an active PR before.

imcotton:
I just realized only need to change from `export =` to `export default` in `localForage.d.ts` file as:

```typescript
declare module "localforage" {
    let localforage: LocalForage;
    export default localforage;  //  from export = localforage;
}
```

and TypeScript will recognize its typing right away.

thgreasi:
@imcotton as far as I know, this wasn't correct in v1.8 since lcoalforage is a UMD module.
Before changing to the `export =` syntax, in my ionic 2 beta project, Typescript recognized it but I was getting errors on runtime. I had to do alchemies like import * as and then cast to the LocalForage type, to get it work.
PouchDB was also using `export default` and now uses `export =` as well (once again, kudos to them).

imcotton:
Odd, since `export default` would gives you correct typing info, how come changed from `export =` in **d.ts** file would cause runtime behavior difference?

thgreasi:
`export default` just declares that it's an ES6 module with an export, BUT localforage is NOT an ES6 module, it is a UMD. No matter what we defined in the d.ts file and how beautifully the TS compiler detects that typings. we will end with runtime errors since localforage doesn't have a `default` export with the ES6 semantics.

Here are two working variations [1](https://github.com/thgreasi/localForage-cordovaSQLiteDriver-TestIonic2App/blob/1412b40ca73231d637dc18bc3d9d1c827936550f/app/localforageCordovasqlitedriverDemo.ts)  [2](https://github.com/thgreasi/localForage-cordovaSQLiteDriver-TestIonic2App/blob/fe12367949a860a1beb26f63a17d03e96bdbea4a/app/localforageCordovasqlitedriverDemo.ts) (TS compliled & runtime worked) back when the d.ts was using the `export default` syntax. `allowSyntheticDefaultImports` could be a solution, but requiring a non-default compilation flag just to compile a specific library cause trouble to big projects that are using multiple libraries.

danbucholtz:
I am one of the tools guys for Ionic. Anything you want me to test out or weigh in on?

Thanks,
Dan

thgreasi:
Ooops closed this pr by accident on my mobile...
@danbucholtz what do you think about this PR and the `export =` syntax that our typings use?
If you are fine with this pr we can merge asap and focus on including the typings as part of the repo.

danbucholtz:
@thgreasi,

I'll clone and test.

Thanks,
Dan

danbucholtz:
@thgreasi,

If I run `grunt build`, I am met with the following warning:

```
Running "browserify:main" (browserify) task
>> Bundle dist/localforage.js created.

Running "browserify:no_promises" (browserify) task
>> Bundle dist/localforage.nopromises.js created.

Running "run:derequire" (run) task
/bin/sh: derequire: command not found
Warning: non-zero exit code 127 Use --force to continue.

```

Am I misunderstanding the build process? I do have files generated in `dist`.

Thanks,
Dan

tofumatt:
This is an issue with us assuming certain tools are globally installed. We
should be looking in node_modules/.bin/derequire

The short-term fix is to npm install -g derequire, but we should change the
path we check for the binary to fix this.

iamolivinius:
npm install -g rimraf derequire ncp

danbucholtz:
This looks good to me:

```
declare module "localforage" {
    let localforage: LocalForage;
    export default localforage;
}
```

This will enable it to work with Ionic and other TS tools without issue. It will use the standard es2015 default export syntax so it will be available to be imported like this:

```
import localforage from 'localforage'
```

Thanks,
Dan

iamolivinius:
The compiler does not generate different code if you make changes to your d.ts file only. But the compiler probably would throw different warning or error messages.

Because localforage is distributed as a UMD module and we're talking about Typescript modules we only need to concentrate on commonjs or amd.

The code below shows several ways of how you can "import" localforage into Typescript(1.8 & 2.0) modules. The code compiles without errors nor warnings. For TS@2.0 I used `@types`, `typings` for TS@1.8.

```ts
// main.ts
// with "allowSyntheticDefaultImports": true in tsconfig.json

declare var require: any

import * as localforage from 'localforage'
import lf from 'localforage'
import forage = require('localforage')
import {setItem} from 'localforage'
const lfrequired: LocalForage = require('localforage')

localforage.setItem("key", "* as")
lf.setItem("key", "default")
forage.setItem("key", "import=require")
setItem("key", "named import")
lfrequired.setItem("key", "commonjs require")
```

```js
// main.js
"use strict";
const localforage = require('localforage');
const localforage_1 = require('localforage');
const forage = require('localforage');
const localforage_2 = require('localforage');
const lfrequired = require('localforage');
localforage.setItem("key", "* as");
localforage_1.default.setItem("key", "default"); // TypeError: Cannot read property 'setItem' of undefined                                                                                                                                
forage.setItem("key", "import=require");
localforage_2.setItem("key", "named import");
lfrequired.setItem("key", "commonjs require");
```

When you run main.js in node or browser it will fail with the TypeError mentioned above. The reason for this are the different semantics between commonjs and es2015 modules.

In es2015 default exports can look like this

```ts
// module.ts
export default "hello world"

// main.ts
import text from './module'
console.log(text)
```

compiled to commonjs (amd is mostly the same)

```js
// module.js
"use strict";
Object.defineProperty(exports, "__esModule", { value: true });
exports.default = "hello world";

// main.js
const module_1 = require('./module');
console.log(module_1.default);
```

TS 1.8 & 2.0 as well as Babel since 6.0 transpile default exports in modules this way. Babel pre 6 did it different, see: https://github.com/babel/babel/issues/2212

These are the reasons why I think `allowSyntheticDefaultImports: true` is not compatible with localforage at this point. If you set it to false the compiler screams at you.

@thgreasi I guess what you're basically trying to do in the current state of https://github.com/thgreasi/localForage-cordovaSQLiteDriver-TestIonic2App is to get localforage running with rollup (node-resolve && commonjs), right? I tried it myself and failed. Would it be possible to add the `jsnext:main` field to the package.json?!

Disclaimer: I'm no Typescript Pro so you better double check what I wrote :smiley:

thgreasi:
@iamolivinius :+1: on your feedback.
> I guess what you're basically **trying** to do... I tried it myself and failed.

The [test ionic 2 app repo](https://github.com/thgreasi/localForage-cordovaSQLiteDriver-TestIonic2App) currently woks for me, both with `ionic serve` and `ionic emulate`. Are you facing any issues?

@danbucholtz Your comment was surprising for me, since [ionic-storage](https://github.com/driftyco/ionic-storage/blob/master/src/storage.ts) (and as a result v2 starter project) works fine (using `import LocalForage from 'localforage';`) with [the current localforage @types](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/localForage/localForage.d.ts), which use the `export =` syntax. Do you suggest that we should change the typings?

Are you fine with the doc changes of this PR?
Can we call a vote of +1 -1 to merge this doc PR and work on a new issue about the rest? In my opinion the suggested docs change can provide more insight and be more useful for the reader than the previous one.

danbucholtz:
@thgreasi,

We use the `allowSyntheticDefaultImports` throughout Ionic. We need it in order to import everything as a `default` export since our bundler (Rollup) is very particular. If we were using Webpack or Browserify, this would be a non-issue.

[This](http://stackoverflow.com/questions/29596714/new-es6-syntax-for-importing-commonjs-amd-modules-i-e-import-foo-require/29598404#29598404) thread sheds some light on it a bit.

Thanks,
Dan

thgreasi:
Had the `allowSyntheticDefaultImports` thing and that stackoverflow thread in mind.
Based on that and the typescript docs:
* libs of traditional module systems should continue to be exported and imported like before, especially for TS pre 1.8 users
* TS 1.8+ allows to use ES style imports on non ES6 modules (+ default imports with `allowSyntheticDefaultImports`)
* TS 2.0 adds the `export as namespace` as a better match for UMD modules

But still not getting an answer from anyone whether this is OK or should be changed...

iamolivinius:
As far as I understand it:

If your tsconfig.json `module` target is (es6|es2015) and you're using a module loader/bundler like rollup or browserify you can set `allowSyntheticDefaultImports: true` and import localforage like this: 

```ts
import localforage from 'localforage
```

If you set `allowSyntheticDefaultImports: false` you can't use rollup.

If your `module` target is (umd|commonjs|amd) setting `allowSyntheticDefaultImports: true` would be wrong. You should stick with 

```
import localforage = require('localforage')
```

I think changing the typings to 

```ts
declare module "localforage" {
    let localforage: LocalForage;
    export default localforage;
}
```

would be wrong at this point. Basically because there is no `export.default` in the localforage umd module. See also first bullet point: https://github.com/Microsoft/TypeScript/issues/2242#issuecomment-83694181

imcotton:
Here is my thought:

## First, user have had two things setup:
  - `export = localforag` in **d.ts** _(no need change)_
  - `allowSyntheticDefaultImports` flag on _(as in this PR)_

## situation A
Your tooling needs `TS > ES6 module > bundle (Rollup alike)`
Use this syntax:

    import my_local_forage from 'localforage';

## situation B
Your tooling needs `TS > CommonJS module > bundle (browserify alike)`
Use either one of these syntax:
  0. `import * as my_local_forage from 'localforage';` // _ES6 style_
  0. `import  my_local_forage = require('localforage');` // _ES6 style mix with `require` function_
  0. `const my_local_forage: LocalForage = require('localforage');` // _mostly same as above_


------

To conclusion: the key is how you're going to set --module option, that leads to different tooling and end user syntax.

imcotton:
One more about `allowSyntheticDefaultImports` flag, **Moment.js** also has [notes][10] to turn it on

[10]: http://momentjs.com/docs/#/use-it/typescript/

thgreasi:
I'm merging this for now since it's more useful when compared to the current docs.
For this PR, I'm not trying to suggest in favour of using the non-default `allowSyntheticDefaultImports` setting, but just want to cover all setups/configurations.
Waiting for suggestions in new PRs.
A big thanks to everybody in this discussion.
Learned a lot while elaborating on this.
Wishes from Greece for a great Saturday night (depending on your time-zone).

PR #609, opened 2016-10-07
docs(README):  add notice for compiler options setting for TypeScript

imcotton:
`allowSyntheticDefaultImports` has been [added][default] in TypeScript 1.8, and set to `true` as default for **jsconfig.json** (not **tsconfig.json**) in tag `2.0.5` Microsoft/TypeScript#10895.

This PR would in addition to comply with #608.

/cc @iamolivinius 

[default]: https://github.com/Microsoft/TypeScript/wiki/What%27s-new-in-TypeScript#support-for-default-import-interop-with-systemjs

### DIFF
--- a/README.md
+++ b/README.md
@@ -171,6 +171,17 @@ installed (eg. `npm install --save-dev babel-plugin-system-import-transformer`).
 
 ## TypeScript
 
+First make sure to have `allowSyntheticDefaultImports` set to `true` for
+`compilerOptions` inside of [**tsconfig.json**](https://www.typescriptlang.org/docs/handbook/tsconfig-json.html) file (or `--allowSyntheticDefaultImports` during Command-Line invocation):
+
+```json
+{
+  "compilerOptions": {
+    "allowSyntheticDefaultImports": true
+  }
+}
+```
+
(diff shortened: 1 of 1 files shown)


thgreasi:
Great PR :+1: thanks!
How about combining with the TS2.0 paragraph like this:
```markdown
To import localForage in TypeScript v2.0, where the [`allowSyntheticDefaultImports` compiler option](https://www.typescriptlang.org/docs/handbook/compiler-options.html) is enabled by default ([check your tsconfig.json](https://www.typescriptlang.org/docs/handbook/tsconfig-json.html)) you should use:
```
???

thgreasi:
I don't want this to be unclear/misleading for users of older typescript versions.

thgreasi:
I would prefer targeting the default configuration of each Typescript version.
In case a v1.8 user changes his `allowSyntheticDefaultImports` option, it might cause other used libraries to stop building/working.

So, a statement like
> To import localForage in TypeScript v1.8 & v2.0, make sure to have `allowSyntheticDefaultImports` set to `true` and use...

might cause troubles to some devs.

If you can change this PR during the weekend it would be great and we can merge it into master asap.

Thanks for pointing this out. your detailed descriptions and the degree of your elaboration.

imcotton:
I think we should move change and discussion back to #608 for better alignment, feel free to rephrase over once reached the agreement, no worries.

PR #611, opened 2016-10-17
feat: add TypeScript typings

### LINKED ISSUE #600: Add typescript definitions to the repo

thgreasi:
Resolves #600.
Should not be published as a bugfix release of 1.4.x,
since it will cause duplicate definition errors for TypeScript users that already use @types.
We should probably release this as v1.5.0.

As an extra future idea for TypeScript testing, we could change the API tests instead and possibly use the compilation result on the test runner.

### DIFF
--- a/typings/localforage.d.ts
+++ b/typings/localforage.d.ts
@@ -1,0 +1,111 @@
+
+interface LocalForageOptions {
+    driver?: string | string[];
+
+    name?: string;
+
+    size?: number;
+
+    storeName?: string;
+
+    version?: number;
+
+    description?: string;
+}
+
+interface LocalForageDbMethods {
+    getItem<T>(key: string): Promise<T>;
+    getItem<T>(key: string, callback: (err: any, value: T) => void): void;
+
+    setItem<T>(key: string, value: T): Promise<T>;
+    setItem<T>(key: string, value: T, callback: (err: any, value: T) => void): void;
+
+    removeItem(key: string): Promise<void>;
+    removeItem(key: string, callback: (err: any) => void): void;
+
+    clear(): Promise<void>;
+    clear(callback: (err: any) => void): void;
+
+    length(): Promise<number>;
+    length(callback: (err: any, numberOfKeys: number) => void): void;
+
+    key(keyIndex: number): Promise<string>;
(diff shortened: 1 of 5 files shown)


thgreasi:
@whitecolor, @iamolivinius, @imcotton waiting for feedback.
@mlynch, @danbucholtz I'm planning to publish a new version of `localforage-cordovasqlitedriver` (that drops @types in favour of the embedded typings) as soon as this gets published.

thgreasi:
@iamolivinius thanks, did a rebase and added that.

typings/localforage.d.ts
```diff
@@ -0,0 +1,108 @@
+
+interface LocalForageOptions {
+    driver?: string | LocalForageDriver | LocalForageDriver[];
```
imcotton: I think it should be `driver?: string | string[];`

thgreasi:
Nice catch! Interestingly enough, that's part of definitely typed long enough and nobody spotted that. Kudos

thgreasi:
@imcotton changed that, thanks!

thgreasi:
@tofumatt are you ok with merging this and schedule it for the v1.5 release?

tofumatt:
Looks good to me, like you said, let's ship it in the 1.5 release! 👍

dagatsoin:
Hi @thgreasi, so for using localforage + sqli driver which branch of localforage and sqli driver do I need to use?
(I get strange behavior and a lot of typescript errors in webstorm. So I would like to try to bump the version)

thgreasi:
Hi there @dagatsoin 
Since you mentioned `typescript errors`, in [TestIonic2App](https://github.com/thgreasi/localForage-cordovaSQLiteDriver-TestIonic2App) I'm using the current latest versions:
```json
"localforage": "^1.4.2",
"localforage-cordovasqlitedriver": "^1.5.0",
```
Keep in mind that Ionic2:
* has [enableSyntheticDefaultImports](https://github.com/thgreasi/localForage-cordovaSQLiteDriver-TestIonic2App/blob/master/tsconfig.json#L3) on by default.
* expects you to use Typescript v2+ for the `@types` imports to work.

PR #612, opened 2016-10-17
Upgrade grunt-contrib-jshint package to 1.0.0

iamolivinius:
Had this issue when running `grunt test`. After upgrading to 1.0.0 the the error went away and grunt executed jobs again.

```bash
Running "jshint:source" (jshint) task                                                                                                                                                 
Warning: Path must be a string. Received null Use --force to continue.                                                                                                                
                                                                                                                                                                                      
Aborted due to warnings. 
```


tofumatt:
Looks good, but this upgrade introduces some new rules. Can you fix the rules and make sure to run `npm test` locally to make sure it passes? Thanks 👍

iamolivinius:
What do you mean by "fixing the rules"? Drop the new rules or fix the parts that were highlighted by the new rules?

tofumatt:
I would fix the errors pointed out by the new rules, as I agree with them.

In the case of the loop though, feel free to comment that rule out for that
line if it’s annoying to fix. I’m not sure :-)

tofumatt:
Sweet, that makes the code better too. Thanks! 👍

PR #613, opened 2016-10-17
Remove grunt task: run

iamolivinius:
As you can see here: https://github.com/localForage/localForage/pull/608#issuecomment-252993609 running grunt is unnecessary hard.

This PR removes the dependency of global installed npm packages: rimraf, ncp, and derequire.


tofumatt:
Sounds good, but this doesn't pass the tests. Could you change the tests as well so they work without these global modules? Thanks! 😄

Issue #596, opened 2016-07-24
Create dist/ during Travis build step and remove from repo


tofumatt:
It's quite annoying having to check in `dist/` artefacts, and now that we use Travis to build for NPM, we should be able to just send the `dist` folder to NPM (and perhaps the `gh-pages` branch as well without checking it into git. This is something I'd like a lot.

It messes with bower uses, but frankly I'd like to drop bower support and focus on just NPM.

nolanlawson:
One easy way to do this is to add `dist` to `.gitignore`, but then put a `"files"` member in `package.json` pointing to a list of the files/folders you want to ship to npm, including `dist`. (A tool like `pkgfiles` is very handy to see which files will be uploaded to npm.) Then you can use http://npmcdn.com/ 's auto-Bower feature to satisfy Bower users.

I've been moving to this pattern in most of my projects because managing `dist` files in Git is such a pain.

PR #614, opened 2016-10-18
Removes dist folder from repository.

iamolivinius:
As mentioned in #596 this PR removes the dist/ folder with generated bundles from the repository. 
Only files listed in `package.json` > `files` will be published to the npm registry (as well as the LICENCE file).

This will break bower support for now. I deliberately did not consider npmcdn's bower.zip feature because it will be removed in Jan 2017.


tofumatt:
Actually, I think we should wait on 2.0 for this change just because it could break things for certain people, but I'll get it merged soon.

Should this wait for 2.0 or is it safe to merge now, @thgreasi?

thgreasi:
That's fine with me.
I guess that this would affect people who npm install the git repo directly.

tofumatt:
Yeah. I'm partially assuming they'd have a build system or something anyway, I wish I had numbers on them but sadly I don't 😢

tylergraf:
So what do we do if we install using `bower`?

tofumatt:
> So what do we do if we install using bower?

Honestly bower support is something I think we've chatted about removing and I've certainly thought of. They have stats like npm (see: https://github.com/bower/registry/issues/39) so I don't know how many people use it, but anecdotally npm usage amongst frontend developers seems to be the norm these days. I'd like our npm support to be better; I am less interested in bower.

That said I would be okay waiting on 2.0 to merge this with a notice that 2.0 would drop bower support.

tylergraf:
Well I definitely wouldn't support removing bower support. I can work around it and add a build step in my process, but I hope you get real numbers before this happens.

tofumatt:
That’s the problem though :-( (see the issue I linked to)

There are no numbers on bower installs short of asking GitHub. I guess I can try it as a one-off though; if they’re way high then you’re right: I won’t nix it (though it’s annoying to maintain when we want to keep dist/ out of the repo).

thgreasi:
+1 for a build step. That's what we are doing in angular-ui-sortable.
Also Polymer/webcomponent guys seem to be in favor of Bower as well.
-- 
*Thodoris Greasidis*

PR #625, opened 2016-11-21
Ensure that 'version' is a number when calling 'localforage.config'

### LINKED ISSUE #304: make `localforage.config` fail if `version` is not a number

rparrapy:
Fixes #304


thgreasi:
:+1:

tofumatt:
This is good though a breaking change. That said, I think we should be strict about accepting arguments, so this works for me.

We'll need to note that this is breaking.

Thanks!

Issue #627, opened 2016-11-23
Throws on server-side in SSR applications


swernerx:
When using the library for redux-persist in my local application I have issues. The problem seems to be that it throws - at import level - when being referenced in any file - even if not used at all. This is problematic especially for applications which use server-side-rendering (SSR) and share common code parts like the redux initialization. With ES2015 in our code it's not even possible to import the module conditionally.

thgreasi:
Even though this is a library primarily focused to be used on the client, can you provide any further info about the error that you are receiving?

thgreasi:
Any feedback on this?

thgreasi:
Any chance this might be related to the situation described in #626 ?

thgreasi:
Closing this for now because of inactivity. Feel free to reopen with extra info or follow up in an other issue.

thgreasi:
Fixed by #633 planned to be released as v1.5

Issue #626, opened 2016-11-21
execution on import


plexigras:
some parts of the code run on import making delayed or conditional usage of localForage impossible.

https://github.com/localForage/localForage/blob/master/src/localforage.js#L55-L63

thgreasi:
Thesev lines are just environment capability checks. Some initialization has to be done so that the default localforage instance is made available and ready to use right away, which is one of the features that localforage offers. Can you give an example of a use case that this initialization is causes limitations?

plexigras:
its quite specific but isomorphic/universal javascript which runs on server as well as browser

thgreasi:
I still don't get your use case exactly. Keep in mind that in isomorphic/universal javascript the object instances are not transferred to the client but are reinstantiated and reinitialized on the client and as a result the driver support detection variables will end up with different values depending on the current environment. Localforage will detect the environment capabilities even if it's on the server. Since localforage, is focused to run on the client none of the included drivers run on nodejs but you can try polyfilling any of them or use a custom driver like https://github.com/localForage/localForage-memoryStorageDriver .
Obviously, rendering a list of "things" on the server, based on data saved on the client is a complex feat that includes an architecture that replicates the user data on the server (which on itself has timing and syncronization issues that need to be addressed). If you can give a brief description of the key pieces of your use case, I might be able to give you better suggestions.

plexigras:
i want a script that can run in the browser as well as in node which means i need some parts of the script to be conditional including localForage which i want to only use in the browser

thgreasi:
> which means i need some parts of the script to be conditional including localForage which i want to only use in the browser

Ok then just use it in the browser.
1. Why is the driver support initialization troubling you? Just wrap your code with the condition check that you mentioned or write a proxy. The included localforage drivers will reject on the server except if you provide a polyfill or a custom driver to run on the server, like the memory storage driver I mentioned earlier. In case you want the exact same to run everywhere, you can use it and prepopulate it with the user's data or write a custom driver to retrieve it from your storage.
2. Is there anything that this library can do for this issue so that we can resolve it?

plexigras:
you are right having the driver checks just run might actually be a feasible solution.

i'm getting some warning when doing so:

```
[0] webpack building...
[0] ./~/localforage/dist/localforage.js
[0] Critical dependencies:
[0] 7:484-491 This seems to be a pre-built javascript file. Though this is possible, it's not recommended. Try to require the original source to get better results.
[...]

PR #633, opened 2016-12-16
fix(localforage): catch constructor's setDriver Promise

### LINKED ISSUE #631: LocalForage generates a rejected promise when imported in node.js
All I need to do is `import localforage from 'localforage'` in a node.js package to generate an error (no available storage found) - I don't have to config localforage or anything. The error itself is in a rejected, and uncaught, promise. The problem is that in future Node.js versions, a rejected promise may end up in the process terminating with a nonzero exit code.

My situation in particular is that I'm not actually using localforage, I'm just importing a package that can be configured to use it. With ES6 style import/export code, it's unclear how one would conditionally import stuff (vs require-style imports, which can happen inside an if-else block).

Ideally, the config stuff that throws this error would be moved into an initialize method or something, instead of [returning a new LocalForage()](https://github.com/localForage/localForage/blob/2f887aa69fcea10d409fc755c208ae0363a64190/src/localforage.js#L383) from the export command.

thgreasi:
Resolves #631 #626 #627.
@tofumatt I'm not sure how I could add tests for this (since it's a warning atm) but if you are fine with it we can merge it.


ericeslinger:
I can confirm that this causes localforage to no longer generate an unhandled promise rejection on require() in a node environment.

thgreasi:
@tofumatt gogo approval?

src/localforage.js
```diff
@@ -116,7 +116,7 @@ class LocalForage {
         this._dbInfo = null;
 
         this._wrapLibraryMethodsWithReady();
-        this.setDriver(this._config.driver);
+        this.setDriver(this._config.driver).catch(() => {});
```
tofumatt: We could probably at least `console.error` this or something if `console.error` exists, just so we don't entirely fail silently (I could see that confusing people in the future).

thgreasi: I was thinking of that, but they are going to get the error at the moment they try to use any localforage method (the "no driver tests" still pass).
So for the developer:
* nothing actually changes with this PR
* localForage seems to do nothing on import, since the error during the default instance initialization is silenced and appears only when a method is used. After 3 duplicate tickets on this, I'm now starting thinking of this as a benefit.

tofumatt:
Haha fair enough actually, I guess we can leave it out for now. 👍🏻

- Matthew Riley MacPherson | tofumatt.com

jasan-s:
I am still getting an error : `(node:12884) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 6): Error: No available storage method found.` I'm using` v1.50`

tofumatt:
@jasan-s I don't think that's related to this pull request. Could you please file a new issue?

Issue #632, opened 2016-12-15
Exclude Promise polyfill


idiotWu:
By now(v1.4.3), this lib will try to import [lie](https://github.com/calvinmetcalf/lie) if there's no `Promise` exists. However, there are several disadvantages of introducing a polyfill in such a third-party library.

1. The main entry is bundled with polyfill.
   
    Though there is a `nopromises` bundle, as long as the one with polyfill is marked as the default entry, most of us will still import `dist/localforage.js`, and thus, our own polyfills may be broken.

2. `nopromises` bundle is not as clean as it's expected to be.

    When using non-automatic Promise polyfills, we're possibly writing codes like:

    ```js
    import Promise from 'promise-polyfill'; 
    import localForage from 'localforage/dist/localforage.nopromises';

    if (!self.Promise) {
      self.Promise = Promise;
    }
    ```

   Since localforage requires Promise as soon as it is executed, we will see a `Cannot find module 'lie/polyfill'` error in our console. That's really confusing me because **I've already been importing the nopromises version**.

AFAIC, I think it's better to strip the Promise polyfill and give users a warning like "*You will need a Promise polyfill for older browsers*." Or, at least, can we wipe out all the polyfill importing sentences in `nopromises` version?

thgreasi:
@tofumatt
In my opinion we should not import/require polyfills in any app or lib for
many reasons, one of which is the aforementioned code clutter. I expect the
environment to be ready. I would suggest to grunt-contat lie just before
the lib's code in the main dist file. Interestingly enough, in my tests
this change lead to a small increase to the resulting bundle size.

> --
*Thodoris Greasidis*

tofumatt:
Yeah, I now agree. At the time when we started localForage Promises support was a bit spotty but now I think we can remove the polyfill.

thgreasi:
> now I think we can remove the polyfill.

1) Oh, pause. You mean remove it completely from the main dist file?
2) how about droping the nopromises file and adding a new one localforage.withpolyfills.js (or something)?
3) may I push it even further and ask to also use some es5 Array methods (and add polyfills to the respective polyfilled build target)?

tofumatt:
1. Yeah I was thinking remove it entirely? I guess we could put it in just the dist file though, fair enough. [Browser support](http://caniuse.com/#feat=promises) seems good enough.
2. I like that
3. Totally, we are already using ES6, right?

idiotWu:
[...]

Issue #615, opened 2016-10-23
Give a note to ionic-storage guys releasing v1.5.0


thgreasi:
Related to #611 
Give a note:
- [x] before the release - driftyco/ionic-storage#28
- [x] after the release - driftyco/ionic-storage#28

thgreasi:
Just opened driftyco/ionic-storage#56 that uses `~` instead of `^` to limit the dependency version of localforage in ionic-storage. As soon as they merge it and release a new version, lfv1.5 will not cause any issues.

thgreasi:
The ionic guys responded and have limited their localforage dependency to avoid typing conflicts.

PR #636, opened 2016-12-19
Return config set driver promise

### LINKED ISSUE #634: Possibly unhandled promise in .config()
@tofumatt now that I started looking for unhandled promises, I just came by [this one in `config()`](https://github.com/localForage/localForage/blob/4fe872cf0f866cb5bcc54848ecbdfb44915bfd08/src/localforage.js#L153).
What do you think we should do? Should we:
* silence it with a `catch()` like we did with the constructor
* return that promise if a `driver` is passed (which is still truthy, so it's probably not breaking)
* other thing that I couldn't think of since its early in the morning

thgreasi:
Resolves #634


thgreasi:
@tofumatt Please review and merge. It's more or less what we talked about in #634 .

tofumatt:
I guess this fixes up a lot of our promise things, should we release 1.5 now to solve them for folks?

thgreasi:
How about also doing #632 first? I should also give a note to the Ionic folks since they might need to resolve some conflicts, now that we include typings.

thgreasi:
@tofumatt we can cut a release. Should we schedule it on 22 Dec, so that we can give a notice (and some time) to the Ionic guys (as per #615)?

tofumatt:
Sounds good, or at this point either the 23 or whenever works for them!

I'll be around and can make one, not too many holiday plans 😄

PR #639, opened 2016-12-30
Update documentation about using with webpack

### LINKED ISSUE #463: Remove the extra instructions for webpack from README
Now that we are targeting the dist files for all bundlers we should remove the extra webpack instructions that suggest using the script loader plugin.
Our support should no longer be considered spotty

stephanebachelier:
Close #463

### DIFF
--- a/README.md
+++ b/README.md
@@ -164,10 +164,13 @@ localForage 1.3+ works with both Browserify and Webpack. If you're using an
 earlier version of localForage and are having issues with Browserify or
 Webpack, please upgrade to 1.3.0 or above.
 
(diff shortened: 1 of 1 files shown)


thgreasi:
I feel shamed that #463 is still open...
Thanks for the contribution.
Merging right away.
My wishes for a happy new year.

stevemao:
adding `noParse` option is just like turning off the fire alarm in a fire. It doesn't help webpack to optimise the build result. Should `main` be the src entry instead (or the entry of a es5 version transpiled by babel)?

PR #643, opened 2017-01-14
fix(indexeddb): fix blob support check

### LINKED ISSUE #618: 1.4.3 always fails blob support test
In localforage 1.4.3 (testing in Chrome) it always fails `_checkBlobSupportWithoutCaching` because it passes a database instead of a transaction, which leads to this call:

`txn.objectStore(DETECT_BLOB_SUPPORT_STORE).put(blob, 'key');`

which throws `TypeError: txn.objectStore is not a function` because `txn` is really an `IDBDatabase`. It is clearly passing a database up the callstack from `_checkBlobSupport(idb)`, initiated by a `setItem` call with a blob type (`_checkBlobSupport(dbInfo.db)`).

Execution jumps down in to a promise `catch` which returns false, and it ends up saving an encoded blob rather than a normal blob directly.

localforage 1.4.0 does not appear to be affected so this seems to be a recent regression.

Demo: https://www.scirra.com/labs/bugs/localforage-blob/
That basically just does `localForage.setItem("foo", new Blob(["bar"]))`. Open dev tools and set pause on all exceptions.

thgreasi:
Fixes #618


thgreasi:
Sorry I didn't have the time to write some tests :disappointed:

thgreasi:
It ends up that we already have tests for blobs which were failing on FF &chrome but are skipped in phantomjs because. This PR fixes the tests in both FF and chrome.

tofumatt:
Sweet! I’ll have a look at a lot of this stuff soon, just a bit swamped
with add-ons stuff right now but I have all the LF queued up for the
weekend :-)

PR #645, opened 2017-01-18
Updated README.md

atanasbozhkov:
Fixed Backbone driver hyperlink.


tofumatt:
Oh, good catch. Really I should move that repo into this organisation. I'll do that instead, but thanks for pointing this out!

Issue #70, opened 2014-02-18
InvalidStateError in Firefox


coderaiser:
In `FF v27.0.1` occurs error:

```
asyncStorage: can't open database: InvalidStateError
InvalidStateError
```

in this code:

``` js
var openreq = indexedDB.open(DBNAME, DBVERSION);
```

In `Chrome v32.0.1700.107 m` everything works fine. 
Could you help me?

tofumatt:
Looks like #448 should fix this one, but if not please let us know and reopen the issue. :smile:

karol-f:
I'm still experiencing 'invalidState' error - [Firefox in 'private mode'](https://www.dropbox.com/s/egknaq8r3bai5or/private_mode_fx.jpg?dl=0)

Do you think @tofumatt that we should reopen this issue?

thgreasi:
@karol-f what version of localforage are you using?
Trying [this codepen](http://codepen.io/thgreasi/pen/ojYKeE) in private mode on Firefox dev edition 48.0a2 fallbacks to the localstorage driver and succesfully retrieves the stored value.

karol-f:
Hi @thgreasi, I'm using latest (1.4.2). Your CodePen is indeed working - [Private mode codepen](https://www.dropbox.com/s/8aihkdufm7hzbfo/private_mode_fx_codepen.png?dl=0) 

I'll check my code for race conditions and misconfigurations (I'm not setting drivers as you in Codepen, I'm not setting them at all).

Thank you for the response!

thgreasi:
This is an example pen and we are only setting the driver preference order to ease experimentation.
Commenting out the `setDriver()` does not affect the result, since this is exactly what localforage does under the hood by default. Keep in mind that setting indexeddb as the only available driver will break the pen on Firefox Private Mode, since as you saw only localstorage is available on that environment.

elyobo:
@thgreasi your pen is exposing an InvalidStateError _and_ also successfully writing and reading via localForage for me with FF 50.1.0+build2-0ubuntu0.16.04.1 in private browsing mode.

![ff](https://cloud.githubusercontent.com/assets/407806/21908285/3ba1005e-d967-11e6-8e94-4124b49defd7.png)

The order of the error is odd though; reading and writing happens fine and then afterwards the error is thrown (which is what I'm seeing in my own application).

elyobo:
[...]

PR #648, opened 2017-01-22
fix(IndexedDB): prevent InvalidStateError on failed open() request

### LINKED ISSUE #647: Changing Firefox settings causes uncatchable exceptions
In Firefox, go to Options -> Privacy -> History and set Firefox to "Never remember history".

Now all localforage calls throw uncatchable exceptions. Demo URL: https://www.scirra.com/labs/bugs/localforage-crash/

This calls `localforage.getItem()` inside a try-catch block and it still throws an unhandled exception. Presumably there's an internal callback that ends up throwing an exception.

IMO this behavior of Firefox is really annoying, since it breaks pages and libraries that haven't been tested specifically with this particular Firefox setting. It also results in bug reports in our software (https://www.scirra.com/forum/localstorage-crash-with-firefox-no-history_t186255). It's not clear what we can do about this simply as a consumer of the library, because there's no way to catch the exception that is thrown. So I think libraries like localforage ought to handle this to hide these quirks from library users.

This is also precisely why I think there ought to be a memory fallback as proposed in PR #555.

thgreasi:
Fixes #647.
Related to #70.


thgreasi:
@tofumatt please merge this if you are fine with it.

thgreasi:
[Here is a codepen using the code from this PR(http://codepen.io/thgreasi/pen/ZLyzqr).
Seems to work on Chrome (+ Private Browsing) and the Error no longer shows up on FF Privare Browsing. Also works in Edge Private and IE11 after replacing the ES6 parts.

PR #652, opened 2017-01-25
Iamolivinius remove grunt run task

tofumatt:
Tweaks to the PR to run the typescript tests via grunt. Closes #613 by implementing it but adding a fix for our new typescript type check run script.

@thgreasi Can you make sure this is okay? The typescript tests don't output anything on my local machine so I can't tell if it's working right, but seems fine.


thgreasi:
Changing the type of a test to make it fail didn't broke `grunt test`. So I guess that we are missing something for now :-/

thgreasi:
Pushing a commit in a minute!

tofumatt:
Cool, thanks! I don't know much about the TS stuff so appreciate the help 😄

PR #657, opened 2017-02-02
Use node 6 on Travis

michielbdejong:
Avoids this error:

````
/home/travis/build/localForage/localForage/node_modules/grunt-rollup/node_modules/rollup/dist/rollup.js:62
	for ( var item of map.entries() ) {
	               ^^
Loading "rollup.js" tasks...ERROR
>> SyntaxError: Unexpected identifier

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ before_install:
 - bower install
 language: node_js
 node_js:
-- '0.10'
+- '6'
 env:
   global:
     secure: LSmHs32nmcjgp3/5hQASgIPUDF/zexvwV/XASRYCDkwJRllDZ8n8wIvHx+D1QVcotx7jSAVEKM2tcfMq/JLXVHUmNNpeG9maUMfZAHA7xFlAo5GNAG3p/rqYe3tan9Xum+Is6A71kXTVnaVyOTG5ApEyj8QUwPfHpv1ziGJu72Y=
--- a/.gitignore
+++ b/.gitignore
@@ -35,3 +35,6 @@ build/
 .yardoc
 _yardoc
 doc/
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by michielbdejong** [comment]
Ran into this, here: https://travis-ci.org/localForage/localForage/builds/197614901

**Comment by michielbdejong** [comment]
Actually, now also saw this:

````
typescript@2.1.5: wanted: {"node":">=4.2.0"} (current: {"node":"0.12.18","npm":"2.15.11"})

So maybe update straight to 4.2? Is there a reason that the build process is still on node 0.10?

**Comment by thgreasi** [comment]
I'm fine with it. @tofumatt ?

**Comment by thgreasi** [comment]
This also somewhat closes #585.

**Comment by michielbdejong** [comment]
Updated straight to 6.9.5; the resulting dist folder seems to be the same for 0.10, 0.12, and 6.0, but different from the one that's currently in master. Also, `npm test` keeps creating a folder `.tscache`, is that related to typescript? I added two small bonus commits to this PR, to address both those issues.

File: .travis.yml [hunk 1]
```diff
@@ -3,7 +3,7 @@ before_install:
 - bower install
 language: node_js
 node_js:
-- '0.10'
+- '6.9.5'
```
thgreasi: How about just setting it to `6` so that we get updates?

thgreasi:
Any feedback on the rebate thing to remove the dist files?

michielbdejong:
Will try to find some time in the weekend!

michielbdejong:
Removed the commit that builds dist and rebased on latest master

thgreasi:
Thanks that's great!

PR #661, opened 2017-02-15
Create 1.5.0 release

tylergraf:
This releases `1.5.0`.


tofumatt:
This PR doesn't add notes to Safari users about the change in data storage engines for 1.5, so I will add info to the changelog and README about it.

@thgreasi If you think we should add a `console.info` to Safari users as well let me know. I'm on the fence about it as it seems overkill, but if someone else says to do it I'm convinced 😄

PR #665, opened 2017-03-01
Attempt to establish a new IDBDatabase connection after InvalidStateError

### LINKED ISSUE #581: Connection closed on Chrome (IndexedDB) in long sessions

```
```

```
0.47%   Chrome 53.0
8.04%   Chrome 49.0
5.91%   Chrome 52.0
85.58%  Chrome 51.0
```

stkao05:
Resolves #581
Hi guys :)

First of all, sorry if these codes look a bit rough. The PR is created just to get some feedbacks from you guys on general idea here. If it sounds, then I will iterate on it more and deliver a more complete version.

__Motivation__
The PR attempt to solve [an issue](https://github.com/localForage/localForage/issues/581) related to `InvalidStateError` on Chrome browser. From production logs, we observe that Chrome's IDBDatabase connection isn't always stable throughout the whole web session; sometimes it just got closed for some unidentified reasons.

Problem with current localforage implementation is that: once the connection encounter `InvalidStateError`, it is basically a dead-end for the apps; there is no way for apps to recover from it except to reload the page again.

While there might be an issue with Chrome's indexedDB, I think it is also reasonable for the client (in this case, localforage) to handle abnormal connection state error. 

__How it works__
The codes here try to re-establish a new connection when `InvalidStateError` occur. Here is the example scenario:
- User calls `localforage.setItem()`.
- DB connection is somehow broken and being closed. `InvalidSateError` occurred when trying to open a transaction on a closed connection.
- localforage return a rejected promise for the `setItem` call.  At the same time, try to establish new connection. localforage readiness is deferred while connection is being established. 
- If connection is successfully re-established, db's readiness is advanced. If connection fails to establish again, db's readiness is rejected.

__Discussion__
The current code retries only once, and if it fails, it will not retry again and the localforage will be unusable. Also, the code only does not retry immediately. The db call will be first rejected, and retry will be carried out later. We can have a more subtle retry strategy
- Retry immediately, so that it is possible to recover the previously failed db call. I have created [another PR](https://github.com/localForage/localForage/pull/666) to demo this. 
- Retry multiple time. We can retry re-establish connection every time db call need to be made, or we can set max number of retry attempt. 

These suggestion might require larger codes change, so would love to discuss with you before implementing these :)

### DIFF
--- a/src/drivers/indexeddb.js
+++ b/src/drivers/indexeddb.js
@@ -115,6 +115,19 @@ function _advanceReadiness(dbInfo) {
     }
 }
 
+function _rejectReadiness(dbInfo, err) {
+    var dbContext = dbContexts[dbInfo.name];
+
+    // Dequeue a deferred operation.
+    var deferredOperation = dbContext.deferredOperations.pop();
+
+    // Reject its promise (which is part of the database readiness
+    // chain of promises).
+    if (deferredOperation) {
+        deferredOperation.reject(err);
+    }
+}
+
 function _getConnection(dbInfo, upgradeNeeded) {
     return new Promise(function(resolve, reject) {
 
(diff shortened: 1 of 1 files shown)


thgreasi:
Thanks for this PR. I will try to review during the weekend. In the mean time, could you please try fixing the code styling errors that Travis CI complains about?

PR #666, opened 2017-03-02
[Version 2] Attempt to establish a new IDBDatabase connection after InvalidStateError

### LINKED ISSUE #581: Connection closed on Chrome (IndexedDB) in long sessions

stkao05:
Resolves #581
Another variation of connection retry. See this [PR](https://github.com/localForage/localForage/pull/665) for more detail

### DIFF
--- a/src/drivers/indexeddb.js
+++ b/src/drivers/indexeddb.js
@@ -115,6 +115,19 @@ function _advanceReadiness(dbInfo) {
     }
 }
 
+function _rejectReadiness(dbInfo, err) {
+    var dbContext = dbContexts[dbInfo.name];
+
+    // Dequeue a deferred operation.
+    var deferredOperation = dbContext.deferredOperations.pop();
+
+    // Reject its promise (which is part of the database readiness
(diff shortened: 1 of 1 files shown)


thgreasi:
Hi there and sorry for taking me so long to review these PRs.
Do you have a personal preference between the two of them? Did one of the two work better on your tests/app?

thgreasi:
I think that I like this PR more than #665, even though its number makes me feel somewhat uncomfortable.
Especially the part of retrying once.
The fact that `_tryReconnect` returns a Promise makes me feel better up front.

Kudos on this :+1: 
1) Have you tried this code on any project? Do you have any results?
2) It would be great to have a test case for this... I will try writing one, probably by manually manipulating the connection's state.

thgreasi:
Rebased & Merged with #694.
Thanks for your huge contribution on this :+1:

PR #682, opened 2017-04-01
Add modules support

salzhrani:
This is mainly for better `webpack` support. also changed how the promise is polyfilled since webpack + babel take care of it.


thgreasi:
We have also tried generating an es6 dist flavor as well, but then removed it.
We are actually trying to reduce the number of dist files generated and provide as few ways to include LF as possible.

Issue #686, opened 2017-04-11
defineDriver's API introduces the need for ready event


gabrielmaldi:
The docs state that:

> You don’t need to run any init method or wait for any `onready` events.

And this is great.

However, when using a custom driver like this:

```js
localforage.defineDriver(myDriver);

localforage.createInstance({
  driver: myDriver._driver
});
```

I get:

<img width="456" alt="screen shot 2017-04-11 at 7 08 29 pm" src="https://cloud.githubusercontent.com/assets/3728897/24932959/489bfab6-1eea-11e7-885a-12d084b054ce.png">

So, instead, I need to use:

```js
localforage.defineDriver(myDriver).then(function () {
  localforage.createInstance({
    driver: myDriver._driver
  });
});
```

In any nontrivial app, a component will need to call `getItem` (or `setItem`, etc.) on a store; but when using a custom driver, it will first need to wait for the instance to be created. So, we are forced to make the module which creates the stores expose a `ready` event/promise, which components wait for before reading/writing to stores.

Would it be possible to remove this limitation from `defineDriver`'s API and just have it queue up tasks the same way everything else in `localForage` works?

Thanks

thgreasi:
We could define the driver syncronously but we need the result of the `_support()` method of the custom driver, which might be a promise. We complete the driver setup right after the result of the `_support()` method gets resolved. It could be probably feasible to change the code to the setup drivers differently, based on existence (or not) of the `_support()` method and it's return type.

gabrielmaldi:
Thanks for your answer. I opened #692 to work on this.

PR #692, opened 2017-05-02
feat(defineDriver): set up custom drivers syncronously if there's no _support property or it's not a function

gabrielmaldi:
This addresses #686 according to https://github.com/localForage/localForage/issues/686#issuecomment-298235696.


dist/localforage.js
```diff
@@ -343,7 +343,7 @@ if (typeof global.Promise !== 'function') {
 },{"2":2}],4:[function(_dereq_,module,exports){
 'use strict';
 
-var _typeof = typeof Symbol === "function" && typeof Symbol.iterator === "symbol" ? function (obj) { return typeof obj; } : function (obj) { return obj && typeof Symbol === "function" && obj.constructor === Symbol ? "symbol" : typeof obj; };
+var _typeof = typeof Symbol === "function" && typeof Symbol.iterator === "symbol" ? function (obj) { return typeof obj; } : function (obj) { return obj && typeof Symbol === "function" && obj.constructor === Symbol && obj !== Symbol.prototype ? "symbol" : typeof obj; };
```
gabrielmaldi: I don't know why this line is changing. Should this be part of the commit?

thgreasi:
I think that the code should be moved outside the `new Promise(function(){...})`, since according to the spec you can't bet that the constructor function runs synchronously.

dist/localforage.js
thgreasi: If you just did a fresh `npm install` then don't worry about it.
It would be great if the `dist` files were in a different commit, but it's not mandatory.

gabrielmaldi:
I think it's guaranteed that the constructor will run immediately:

From [MDN](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Promise#Parameters) (emphasis mine):

> `new Promise( /* executor */ function(resolve, reject) { ... } );`
> **executor**
> A function that is passed with the arguments resolve and reject. **The executor function is executed immediately by the Promise implementation, passing resolve and reject functions (the executor is called before the Promise constructor even returns the created object).** The resolve and reject functions, when called, resolve or reject the promise, respectively. The executor normally initiates some asynchronous work, and then, once that completes, either calls the resolve function to resolve the promise or else rejects it if an error occurred.
If an error is thrown in the executor function, the promise is rejected. The return value of the executor is ignored.

From [promisejs.org](https://www.promisejs.org/#construction) (emphasis also mine):

> We use new Promise to construct the promise. **We give the constructor a factory function which does the actual work. This function is called immediately with two arguments.** The first argument fulfills the promise and the second argument rejects the promise. Once the operation has completed, we call the appropriate function.

However, if you still feel it's safer to move the code outside the constructor, I'll make the change.

thgreasi:
My bad :sad:
It seems I was a bit confused with an other part of the Promises spec.
If the current code works for you then I;m fine with it.
It would be great to have a test case as well, but if it troubles you then I can give it a try during the weekend,
Kudos for the PR.

dist/localforage.js
gabrielmaldi: Yeah, that's it. I separated the code into two commits 👍

thgreasi:
Thanks for this!!

gabrielmaldi:
Thanks for the tests, I had taken a look and was planning on doing something very similar to what you did. 👍

Issue #620, opened 2016-11-11
SUGGESTION: Add methods to drop instances (databases) and stores (collections)


jlchereau:
Is this hidden somewhere and have I missed it?
Dropping/deleting collections would be really useful, the database is less urgent.

thgreasi:
Is [`clear()`](http://localforage.github.io/localForage/#data-api-clear) too much for your needs?
Does something like [removeItems](https://github.com/localForage/localForage-removeItems) sound better?

jlchereau:
There is a difference between ```DROP DATABASE``, ```DROP TABLE``` and ```DELETE * FROM TABLE``` in SQL. 
If SQL makes this difference, if mongoDB and others make it too, there might be a need for it in localForage, don't you think?

My need is to upgrade the schema without leaving empty useless stores/tables/collections after the upgrade.  Does it  make sense?

thgreasi:
I was 100% with you about the DROP part, I was just asking about the collection data part that you mentioned.
I will do my best to add such a method in v1.5.
I would like to hear some api suggestions. Maybe something like dropInstance() or clearInstance() or deleteInstance(). I guess that we should be able to invoke it:
* without arguments to drop the table of the current instance
* with 2 string arguments defining the table of the db that we need to drop
* with 1 string argument defining the db we need to drop
Any thoughts on these? (cc: @tofumatt)

An extra consideration is whether we should add the new method as a required or optional method of the custom driver api. Making it required would make the new release backwards incompatible.

jlchereau:
I have only ever experienced different flavors of SQL databases and mongoDB, and they both *drop* so I would definitely favor **dropInstance** over clearInstance and deleteInstance. Furthermore clearInstance is ambiguous considering the clear method in localForage.

Considering the existing createInstance:

```
localforage.createInstance({ name: dbName, storeName: storeName });
```

It makes sense to maintain the same method signature (as you suggest):

```
localforage.dropInstance({ name: dbName, storeName: storeName });
```

As far as backward compatibility is concerned, I have no idea how many people have developed drivers for localForage and how easy it is to communicate with them. In my opinion, backward incompatibility is not a big issue if localForage can report its version like jQuery has ```$.fn.jquery```.

thgreasi:
@jlchereau OK, `dropInstance` was all I needed to hear. I will try to take a look at it during the weekened.

thgreasi:
Still working on this. My initial findings include:
[...]

PR #694, opened 2017-05-07
feat(indexeddb): try reconnect to the DB after InvalidStateError

### LINKED ISSUE #581: Connection closed on Chrome (IndexedDB) in long sessions

thgreasi:
Merges #666 
Resolves #581
Closes #665

Build on top of @stkao05 's great PR & idea.
Adds some tests that cause an InvalidStateError by manually closing the db.
Adds some extra checks.
Refactored the code to use callbacks for the async creation of the transaction, since FF disposes the transaction when returning it in a Promise and the object store couldn't be opened.

I :+1:  for this to be merged and waiting for feedback from @stkao05 & @tofumatt.

PS: Keep in mind that the IDB tests do not run on the current (outdated) PhantomJS that we use. I had to `grunt serve` and open `test/test.main.html` in FF and Chrome tp test this.


thgreasi:
After closing manually the IDB connection with `localforage._dbInfo.db.close();`
The following works:
```js
var tx = localforage._dbInfo.db.transaction(localforage._dbInfo.storeName, 'readonly');
(function (tx) {
    console.log(tx);
    var store = tx.objectStore(localforage._dbInfo.storeName);
    console.log(store);

    var req = store.get('STORE_KEY');
    req.onsuccess = function() { console.log(req.result); };
    req.onerror = function() { console.error(req.error); };
})(tx);
```

But these two fail in FF:
```js
// FF doesn't like Promises (micro-tasks) and IDDB transaction/store operations
Promise.resolve(localforage._dbInfo.db.transaction(localforage._dbInfo.storeName, 'readonly')).then(tx => {
    console.log(tx);
    var store = tx.objectStore(localforage._dbInfo.storeName);
    console.log(store);

    var req = store.get('STORE_KEY');
    req.onsuccess = function() { console.log(req.result); };

    req.onerror = function() { console.error(req.error); };
    
}).catch(e => console.error(e));



var tx = localforage._dbInfo.db.transaction(localforage._dbInfo.storeName, 'readonly');
(function (tx) {
    console.log(tx);
    var store = tx.objectStore(localforage._dbInfo.storeName);
    console.log(store);

    Promise.resolve(store).then(function(store) {
        var req = store.get('STORE_KEY');
        req.onsuccess = function() { console.log(req.result); };

        req.onerror = function() { console.error(req.error); };
    }).catch(function(err) {
        console.error(err);
    });
    
})(tx);
```

thgreasi:
This might also be useful for #620 (`dropInstance()`), where we had to close and reopen all the idb connections

thgreasi:

tofumatt:
Hey sorry I've been on holiday this week but will look right now 😅

src/drivers/indexeddb.js
```diff
@@ -399,37 +468,44 @@ function iterate(iterator, callback) {
...
+
+                        if (cursor) {
+                            var value = cursor.value;
+                            if (_isEncodedBlob(value)) {
+                                value = _decodeBlob(value);
+                            }
+                            var result = iterator(value, cursor.key,
+                                iterationNumber++);
+
+                            if (result !== void(0)) {
```
tofumatt: This could probably use a comment just to explain things a bit more for those unfamiliar with the `iterator()` function used here.

src/drivers/indexeddb.js
```diff
@@ -648,28 +764,35 @@ function keys(callback) {
 
     var promise = new Promise(function(resolve, reject) {
         self.ready().then(function() {
-            var dbInfo = self._dbInfo;
-            var store = dbInfo.db.transaction(dbInfo.storeName, 'readonly')
-                .objectStore(dbInfo.storeName);
+            createTransaction(self._dbInfo, 'readonly', function(err, transaction) {
```
tofumatt: I wonder if there is already a constant for `'readonly'` and `'readwrite'` we could use? if not, maybe we can make some rather than repeating the strings everywhere?

src/drivers/indexeddb.js
thgreasi: To be clear, I just changed the indentation here, but this is an opportunity to add some comments as well :)

stkao05:
Hi @thgreasi,
Sorry for the late reply as I was on vocation. I took a look at your revision and they looks great to me 👍

One thing I am worrying about is that - I don't know if IDB connection could be re-establish successfully after InvalidStateError in the real life situation. I will try it out on our production, and let you know if such a fix is helpful. 😉

ccorcos:
Any ideas when this will be officially released?

tofumatt:
Not sure yet but I’d like to take some time to look at localForage this weekend. I know I’m very behind on reviews but unfortunately there’s a lot of add-ons work that’s taking up a lot of time. Thanks for your patience.

ccorcos:
Sounds good. I'm just seeing some InvalidStateErrors. I might just use the latest commit instead of a version

ccorcos:
FYI I've been using this in production at www.notion.so for the past 36 hours and haven't gotten a single InvalidStateError since upgrading to d58233e73f6d5e4878b615a3576fcb4b6428e9b6 (we were seeing hundreds)

tofumatt:
Awesome, that's great to hear! Thanks for letting us know 😄

thgreasi:
@tofumatt Here is a list with the most important changes on master since v1.5.0
* 64c564b feat(typings): added generics to iterate function
* 4db9f4a fix(indexeddb): fix concurrent requests after InvalidStateError
* 1d17b30 fix(indexeddb): try re-establish DB connection using callbacks
* 7f75f07 fix(indexeddb): check db existence before closing
* 557c364 fix(indexeddb): try re-establish IDBDatabase connection after encoutner InvalidStateError error
* 3f7df54 feat(defineDriver): set up custom drivers syncronously if there's no _support property or it's not a function.

How about releasing as v1.5.1 whenever you got some time to run an `npm publish`?
We can then iterate from there with the other PRs and track any regressions easier.

PR #695, opened 2017-05-08
Update URLs in Docs README with new organization

eppfel:


thgreasi:
Thanks! Nice catch :+1:

Issue #688, opened 2017-04-25
Remove instance on IndexedDB


LeonardoCiaccio:
How i can remove this with code ?

![screenshot-jsfiddle net 2017-04-25 14-37-01](https://cloud.githubusercontent.com/assets/1698144/25385576/d25ecf14-29c4-11e7-9ff9-877d1171bc13.png)

thgreasi:
Localforage doesn't provide an API for such operations at the moment.
You should use the WebSQL API directly.
I'm closing this for now and point you to #620 where we are tracking the progress for such an API.
Feel free to re-open in any case.

thgreasi:
@LeonardoCiaccio  The `dropInstance()` PR in in the current master (and planned to be released as LF v1.6) can you give it a try?

PR #708, opened 2017-07-23
feat(localforage): implement dropInstance()

### LINKED ISSUE #620: SUGGESTION: Add methods to drop instances (databases) and stores (collections)

thgreasi:
Resolves #620, #688.
This was very confusing and took ages (aka almost 9 months) to get it right.
@jlchereau, @LeonardoCiaccio  do you mind testing?
Keep in mind that PhantonJS doesn't run the IndexedDB tests, so I had to do `grunt serve` and open the test page on a browser.


jlchereau:
Hi @thgreasi, thank you for progressing this. Unfortunately I won't have any time to commit on testing this new feature before September. Let me know whether this works for you.

thgreasi:
@jlchereau I was super-excited about this and now I'm heartbroken 💔 
Seriously now, no worries.

tofumatt:
I will still have a look at this, though it might take me longer than usual! I'm visiting some friends in Canada and catching up on add-ons stuff, so I might be a week or two. (Same goes for other LF PRs, but I will address them!)

If anything's really urgent let me know and I'll make some time.

jlchereau:
Thanks you for progressing this. At this stage, I have only tried with Chrome 62 on Windows 10, which uses indexedDB by default. 

`localforage.dropInstance({ name: 'db', storeName: 'table' }, function(err) {...});` works fine when both db and table exist and err is null which is consistent with all localForage methods.

`localforage.dropInstance({ name: 'db' }, function(err) {...});` returns err of type IDBVersionChangeEvent  and the database db is not removed which is unexpected.

`localforage.dropInstance({ name: 'dummy', storeName: 'dummy' }, function(err) {...});` where dummy does not exist raises Uncaught TypeError: Cannot read property 'deferredOperations' of undefined which could be handled more nicely.

I hope this helps.

thgreasi:
Thanks for the detailed report.
Could you [fork this codepen example](https://codepen.io/thgreasi/pen/ojYKeE) to demonstrate the error in `localforage.dropInstance({ name: 'db' }, function(err) {...});`?
That's strange since I think that there is a [test case for this](https://github.com/localForage/localForage/pull/708/files#diff-c82a26ff80221916a144cb95e0cdc72dR1526)!?

jlchereau:
Try this:
[lf.zip](https://github.com/localForage/localForage/files/1450205/lf.zip)

Note: to deactivate a test case, replace `it` by `xit`

jlchereau:
I have just realized that trying to dropInstance 'dummy' which supposedly does not exist actually creates `dummy`:

image

jlchereau:
Note: I should have reference `localforage.nopromises.js` instead of `localforage.js` but still some of my findings are still valid.

jlchereau:
Have you been able to reproduce?

thgreasi:
Hey @jlchereau .
First of all **Thanks for providing this feeback as test cases!**
Yea, I saw the timeout. Interestingly enough increasing the timeout to 30secs seems to resolve the issue of dropping the third store.
Didn't have enough time to figure out a solution though.
Hopefully I can find further clues this weekend.

thgreasi:
@jlchereau did a profiling and got the following interesting results:
```
dropInstance:dbPromise: 0.831298828125ms
dropInstance:dropObjectPromise-db.close(): 0.0390625ms
dropInstance:dropObjectPromise-open: 4.756103515625ms
dropInstance:deleteObjectStore(store1): 0.055908203125ms
dropInstance:dropObjectPromise-onupgradeneeded: 8095.27001953125ms
dropInstance:dropObjectPromise-onsuccess: 8116.884765625ms
dropInstance:dropObjectPromise: 8120.800048828125ms
dropInstance:_advanceReadiness: 0.113037109375ms
dropInstance: 8140.351806640625ms
```
For some reason `deleteObjectStore` IDB method call takes ages... I will try to investigate whether there is something that stalls it.

thgreasi:
@jlchereau let's move the discussion to #779 where I will post an update for this in a bit.

PR #713, opened 2017-08-01
Added Generics to iterate function

DaSchTour:
Replacing any with generics to improve typ strength. This way the function that is passes as a parameter will be checked if it returns the expected output and type information from iteratee to callback function are "passed".

### DIFF
--- a/typings/localforage.d.ts
+++ b/typings/localforage.d.ts
@@ -35,9 +35,9 @@ interface LocalForageDbMethods {
     keys(): Promise<string[]>;
     keys(callback: (err: any, keys: string[]) => void): void;
 
-    iterate(iteratee: (value: any, key: string, iterationNumber: number) => any): Promise<any>;
-    iterate(iteratee: (value: any, key: string, iterationNumber: number) => any,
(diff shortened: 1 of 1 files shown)


thgreasi:
What if multiple types are stored in the store?
Should we have to do something like `.iterate<Type1|Type2|Type3, Type1|Type2|Type3>(/*....*/)`?
Should we also keep the `any` interfaces as well and also let the user define the type(s) that they expect with your overloads?

DaSchTour:
Yes that's right. There would be a list of types that could inside the storage.

For TypeScript 2.3 it would be possible to default generics to any. So if nothing is set any is assumed. It is also still possible to explicitly set this to any by using `.iterate<any, any>`

I think for multiple types stored at least the output will be on type in many cases, so definitions will look like this `.iterate<any, Type1>` or `.iterate<Type1|Type2|Type3, Type4>`

But maybe it's also a good idea to keep the old definitions. I'll check if it behaves like I expect.

thgreasi:
Thanks 👍

PR #721, opened 2017-08-14
Add memory storage driver

### LINKED ISSUE #715: Fallback to another storage method when localStorage isn't available
Hello,

We are currently using localForage in our PWA by means of ionic storage, we are really happy :)

We are facing an issue with Safari and localStorage in private browsing mode (as many other people) when trying to save data. It throws an error and it doesn't use any other fallback because it treats the error as if something wrong happened instead of "this storage method is not available". Private navigation in browsers, when it comes to storage, is not a pretty good thing as you already know.

I think that the issue is that localForage's [localStorage supports function](https://github.com/localForage/localForage/blob/master/src/utils/isLocalStorageValid.js), and [localStorage driver's _initStorage function](https://github.com/localForage/localForage/blob/master/src/drivers/localstorage.js#L11) determine that localStorage is available when it is actually not letting save anything. I think that it could be solved by checking if it is actually saving data, and not throwing a quota error, in any of these methods.

It would be really useful when having a localStorage fallback method like in-memory storage because if it is not checked it wouldn't be used at all.

That said, I'd be happy about solving it in a PR if you agree that this is the way to go with this issue.

Thank you!

thgreasi:
Rebases and merges #555 to give credits to @rubenstolk for the original proposal and merges the implementation from [localForage-memoryStorageDriver](https://github.com/localForage/localForage-memoryStorageDriver).
Resolves #715.

The new driver is called `memoryStorage` instead of the `memoryStorageDriver` that the custom driver used. I'm not strongly opinionated on that and would love to get some feedback. Presumably it might prevent naming conflicts and is closer to `asyncStorage` and `webSQLStorage` driver names.

Last but not least, let me add a Woohoo! 🎆

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -303,12 +303,14 @@ By default, localForage selects backend drivers for the datastore in this order:
 1. IndexedDB
 2. WebSQL
 3. localStorage
+4. in memory (does not persist after a page refresh)
 
 If you would like to force usage of a particular driver you can use `setDriver()` with one or more of the following arguments:
 
 * `localforage.INDEXEDDB`
 * `localforage.WEBSQL`
 * `localforage.LOCALSTORAGE`
+* `localforage.MEMORY`
 
 <aside class="notice">
   If the backend you're trying to load isn't available on the user's browser, localForage will continue to use whatever backend driver it was previously using. This means that if you try to force a Gecko browser to use WebSQL, it will fail and continue using IndexedDB.
@@ -348,7 +350,7 @@ Set and persist localForage options. This must be called *before* any other call
   <dt>driver</dt>
   <dd>
     The preferred driver(s) to use. Same format as what is passed to <a href="#settings-api-setdriver"><code>setDriver</code></a>, above.<br>
-    Default: <code>[localforage.INDEXEDDB, localforage.WEBSQL, localforage.LOCALSTORAGE]</code>
+    Default: <code>[localforage.INDEXEDDB, localforage.WEBSQL, localforage.LOCALSTORAGE, localforage.MEMORY]</code>
   </dd>
   <dt>name</dt>
   <dd>


tofumatt:
Sweet! I'll have a look after Tuesday as I'm trying to get a few patches into the add-ons site. Should be good to look Tuesday night or Wednesday morning. 👍

thgreasi:
I think that dropInstance() #708 should go first, since this is a breaking change and should target LF v2.
I'm now working on a refactor for v1.5.x that will provide a way to work around the Safari IDB breaking change.

thgreasi:
@tofumatt Can I also get a review on this?

tofumatt:
Oh shoot, yes! I’m out for dinner and away from my computer tonight but I’ll get to it tomorrow for sure.

- Matt (Sent from mobile)

thgreasi:
Doing a moching a merge commit ATM and will update this pr that to make the comparison a bit easier.

src/drivers/memory.js
```diff
@@ -0,0 +1,234 @@
...
+                delete db[key];
+            }
+        }
+    });
+
+    executeCallback(promise, callback);
+    return promise;
+}
+
+// Retrieve an item from the store. Unlike the original async_storage
```
tofumatt: I don't think a lot of these comments are needed. Some of them seem copy+pasted and not even super-relevant to the in-memory driver.

src/drivers/memory.js
```diff
@@ -0,0 +1,234 @@
...
+}
+
+// Remove all keys from the datastore, effectively destroying all data in
+// the app's key/value store!
+function clear(callback) {
+    var self = this;
+    var promise = self.ready().then(function() {
+        var db = self._dbInfo.db;
+
+        for (var key in db) {
```
tofumatt: For `clear`, couldn't see just set `self._dbInfo.db = {}` rather than manually deleting every key?

src/drivers/memory.js
```diff
@@ -0,0 +1,234 @@
...
+        }
+    });
+
+    executeCallback(promise, callback);
+    return promise;
+}
+
+// Retrieve an item from the store. Unlike the original async_storage
+// library in Gaia, we don't modify return values at all. If a key's value
+// is `undefined`, we pass that value to the callback function.
```
tofumatt: Also it looks like that comment isn't even true anymore 😉

src/drivers/memory.js
```diff
@@ -0,0 +1,234 @@
...
+        if (db.hasOwnProperty(key)) {
+            delete db[key];
+        }
+    });
+
+    executeCallback(promise, callback);
+    return promise;
+}
+
+// Set a key's value and run an optional callback once the value is set.
```
tofumatt: Yeah I think all these comments can go 🔥

src/drivers/memory.js
```diff
@@ -0,0 +1,234 @@
...
+        // Convert undefined values to null.
+        // https://github.com/mozilla/localForage/pull/42
+        if (value === undefined) {
+            value = null;
+        }
+
+        // Save the original value to pass to the callback.
+        var originalValue = value;
+
+        function serializeAsync(value) {
```
tofumatt: This looks like it should be a utility function. Right now it's defined every time `setItem` is called, which seems wasteful.

test/test.nodriver.js
```diff
@@ -55,29 +53,73 @@ describe('When No Drivers Are Available', function() {
...
+
+        // Reset localForage before each test so we can call `config()` without
+        // errors.
+        beforeEach(function() {
+            localforage._ready = null;
+            localforage.config(this.defaultConfig);
+        });
+
+        after(function() {
+            this.defaultConfig.driver = DRIVERS.slice();
```
tofumatt: I'm guessing this is to return a new copy, but could we use `[ …DRIVERS ]` for that instead (I forget if ES6 will let us)? Or even use a `deep-copy` library for the tests?  It just seems more idiomatic than using `.slice()`, which to me is more taking advantage of the side-effects of `slice()` than intending to use it to actually change the array. 😃

src/drivers/memory.js
thgreasi: I think that there was a reason for this at some point, but with the current code that's definitely better :+1:

src/drivers/memory.js
thgreasi: :+1: I will move this outside `setItem` for now but we should definitely refactor the serializer to return promises by default.

test/test.nodriver.js
thgreasi: I will add a comment for now explaining the purpose of this, but we should definitely add es6 support in our tests. Can't even use `() => {}` right now, which seems frustrating. I actually got bitten by this while updating the the test cases.

tofumatt: I have a branch with this in process. I converted the Api tests already to use karma (tests in all browsers), es6, async/await, etc. I’ll get back to it when I’m back from holiday :-)

src/drivers/memory.js
thgreasi: Oh, the reason was that other instances that use the same db will still have access to the old data!
We will have to use a db provider method anyway, so that instances can continue working when their db is dropped.

thgreasi:
Merged with 40c70210a54645c33919a283b2a1138f40b7a001

Issue #676, opened 2017-03-21
localForage 1.5: IndexedDB not available anymore in iOS <10.3 despite installed base due to 1.4.3 bug


denis1994:
# bug: localForage 1.4.3 enables AsyncStorage (IndexedDB) on iOS 10

The default storage changes from WebSQL to AsyncStorage (IndexedDB) on upgrading from iOS 9 to iOS 10 with localForage 1.4.3.

* localForage 1.4.3, iOS 9.3 (tested in iPhone Simulator)

```
	- Default driver: "webSQLStorage"
	- Supported drivers: ["webSQLStorage","localStorageWrapper"]
```

* localForage 1.4.3, iOS 10.2, iPhone 6s Simulator

```
	- Default driver: "asyncStorage"
	- Supported drivers: ["asyncStorage","webSQLStorage","localStorageWrapper"]
```

This seems to be a bug - as this behavior has now been 'fixed' with https://github.com/localForage/localForage/pull/651

## problem: actual users using IndexedDB in iOS 10, 10.1 & 10.2 can't access their data anymore with localForage 1.5

* we have now users in our installed base who actually use this IndexedDB in iOS 10 - below the Safari 10.1 (i.e., without the reworked IndexedDB2.0 in Safari 10.1, which gets released with iOS 10.3).

* However, localForage 1.5 disables using the IndexedDB on iOS <10.3 (Safari <10.1), i.e., we can't access that previously already used storage option

* reverting to 1.4.3 is not an option, as we need to use localForage 1.5 to have the correct behaviour to new users

## request for solution: re-enable IndexedDB for Safari < 10.1 (as option with warnings)
--> Request: re-enable IndexedDB for Safari < 10.1 in localForage 1.5 with an option (and some warning to not use this) 

This would enable us to create a migration back to WebSQL on iOS <10.3 for our installed base users

thgreasi:
According to #604 #650 #651 we added support for IndexedDB in Safari 10.1+ in LF v1.5.
Are you sure that you `used IndexedDB under iOS 10` with `v1.4.3` ?
Also please take a look at the [v1.5 changelog](https://github.com/localForage/localForage/blob/master/CHANGELOG.md).

Lastly, you might want to use [localForage-cordovaSQLiteDriver](https://github.com/thgreasi/localForage-cordovaSQLiteDriver) in your implementation.

denis1994:
I was also surprised that IndexedDB is used.

[...]

PR #723, opened 2017-08-14
feat: use custom driver API to define internal drivers

### LINKED ISSUE #296: Move to using custom driver API to implement our default drivers
It seems there is no reason we shouldn't be using `defineDriver()` ourselves to implement the default localForage drivers. When we check to make sure a user isn't overwriting a driver, we can just check against drivers defined by localForage on start _and_ the user.

I figure this might clean up the code a bit?

thgreasi:
Resolves #296.
Since we now treat internal drivers and custom ones equally, we now allow users to re-define & override internal drivers as well.

This will hopefully also allow LF1.4 migrators (#676) to:
* retrieve the IDB driver
* set the old `_support` check method
* re-define the driver


thgreasi:
cc: @denis1994, @danielsogl, @jesusbotella

jesusbotella:
I think that this will be a huge improvement when it comes to flexibility because we can define the storage drivers the way we need.

I have another question related to this, and the issue #715 that I opened. Will you modify (or be ok with modifying via PR) the localStorage check method?

thgreasi:
I think that this should be part of a separate PR. We can perhaps add run a dummy `setItem` in `_initStorage`, in a similar way to what indexeddb does to detect blob support.

jesusbotella:
Yeah, that is what I wanted when I opened the issue :)

Do you mind if I make a PR to change it in the isLocalStorageValid
function? I saw that you import that file from the utils folder so I
presume there won't be any conflicts :)

jesusbotella:
Sorry, I saw that It would be better in _initStorage as you said 😅

thgreasi:
@jesusbotella I would prefer to keep the validity methods without consequences. I would prefer the test setItem call to be inside the [_initStorage](https://github.com/localForage/localForage/blob/d58233e73f6d5e4878b615a3576fcb4b6428e9b6/src/drivers/localstorage.js#L30) method, in a similar manner with what the indexeddb driver does (having a constant for the test "database" & following a similar naming schema).

We would also need a test case for that, probably extending the [faultydriver test cases](https://github.com/localForage/localForage/blob/master/test/test.faultydriver.html) to override `localStorage` with a method that would throw an error.

thgreasi:
@tofumatt any chance I can get a review on this and possiblty merge and release v1.5.1 during the weekend?

test/test.customdriver.js
```diff
@@ -43,7 +43,7 @@ describe('When Custom Drivers are used', function() {
         });
     });
 
-    it('fails to define a custom driver with overlapping driver name',
+    xit('fails to define a custom driver with overlapping driver name',
```
tofumatt: Should these still be marked as pending? That's what `xit` does, right? Should we remove them or should they be passing?

src/localforage.js
```diff
@@ -46,37 +52,21 @@ var DefaultConfig = {
...
-driverSupport[DriverType.LOCALSTORAGE] = isLocalStorageValid();
-
-var isArray = Array.isArray || function(arg) {
-        return Object.prototype.toString.call(arg) === '[object Array]';
-    };
-
 function callWhenReady(localForageInstance, libraryMethod) {
     localForageInstance[libraryMethod] = function() {
-        var _args = arguments;
+        const _args = arguments;
```
tofumatt: I've been doing ES6 on add-ons and it's nice to see it creeping into localForage 😄

src/localforage.js
```diff
@@ -164,41 +152,37 @@ class LocalForage {
...
                     }
                 }
 
-                var setDriverSupport = function(support) {
-                    driverSupport[driverName] = support;
-                    CustomDrivers[driverName] = driverObject;
+                const setDriverSupport = function(support) {
+                    DriverSupport[driverName] = support;
+                    DefinedDrivers[driverName] = driverObject;
+                    // don't use a then, so that we can define
```
tofumatt: Nice 👍 

Dunno if we did this before but that's clever.

thgreasi: This was done in #692 by @gabrielmaldi, I just added the comment :)

test/test.customdriver.js
thgreasi: I think that we should remove these completely.
We always allowed redefining custom drivers but prohibited redefining internal drivers, possibly without any specific reason in mind.
Allowing to override the internal drivers as well can let us bypass issues like #676.
If you are fine with this, I'm going to rebase, delete the `xit`ed tests and merge this today and you can release this as the next patch release for v1.5 whenever works best for you.

tofumatt: Ah, that makes sense. Yeah, I think we did it originally to make sure users didn’t break anything. But I think allowing them to refine them is a good idea, like you say.

Maybe we could log a console message if they redefine any driver though? That way they at least know what’s up. But yeah, removing these and merging sounds good! 👍🏻

thgreasi: I remember a case that a user complained about some logs on perfectly valid use cases, but we can do that here and check their reactions.
I'm going to use `console.info` if that's ok with you (hopefully later today).

tofumatt: Ah, right. I think it's worth trying... users can always build a version that omits console statements... or maybe we could build our own logger in the future that can be disabled.

Anyway yeah info or even debug is fine for this 👍

thgreasi:
@tofumatt How about releasing the current master as v1.5.1 ?

PR #730, opened 2017-08-21
Check if localstorage is usable

### LINKED ISSUE #729: Check whether localstorage is usable
See: https://github.com/localForage/localForage/issues/715

jesusbotella:
Added code to _initStorage method in order to check if the localStorage API is actually usable in Safari Private Browsing mode. Mentioned in #729 and #715.

What the code does is to save an item (with `dbInfo.keyPrefix + 'check'` as key, and true as value) when the localStorage is being initialized, and remove it if it has been actually added and no error is thrown.

I think that this was the way to solve the issue properly, but what would happen if the user has exceeded the quota by adding items and it is not in private browsing mode? In this case, the localStorage wouldn't be used even though it should be :(

Resolves  #729


src/drivers/localstorage.js
```diff
@@ -24,6 +24,14 @@ function _initStorage(options) {
         dbInfo.keyPrefix += dbInfo.storeName + '/';
     }
 
+    try {
+        var localStorageTestKey = dbInfo.keyPrefix + 'check';
```
thgreasi: How about abstracting this into a new function and using a different prefix like in indexeddb?
eg `_localforage_support_test`

jesusbotella:
I have refactored my changes to abstract the localStorage check into a new function, please tell me if it fits your current naming convention and coding style :)

I have found the way of checking if localStorage is available without a false negative in the case of users that have exceeded the quota without being on private browsing mode. It involves checking the current localStorage items length.

However, it doesn't pass the tests on my local machine because it seems that test.main.html tests save items in the localStorage and don't delete them afterwards, making the length greater than 0. Do I need to worry about it, or it is correct seeing that it doesn't belong to the faulty drivers test?

And at last but not least, what do you mean by "move the dist files in a different commit to ease rebasing"? If I'm not wrong they are in a [different commit](https://github.com/localForage/localForage/pull/730/commits/70edf236f16a75d3bc760031fdb368b8bdbb7375), right? I do not know what to do here 😁

thgreasi:
> It involves checking the current localStorage items length.

I'm not sure whether we should wrap `localStorage.length` inside a `try...catch`. Can you confirm that it doesn't throw in private Safari?

> Do I need to worry about it, or it is correct seeing that it doesn't belong to the faulty drivers test?

How about doing a `localStorage.clear()` inside a `before()` call, like in [here](https://github.com/localForage/localForage/blob/d58233e73f6d5e4878b615a3576fcb4b6428e9b6/test/test.api.js#L163)?

> If I'm not wrong they are in a different commit, right?

You are right, didn't see that 👍

thgreasi:
👍 from me, just add the `clear()` call in the tests and add the new `dist`.
@tofumatt ?

jesusbotella:
> I'm not sure whether we should wrap localStorage.length inside a try...catch. Can you confirm that it doesn't throw in private Safari?

I have just double-checked it and executing `localStorage.length` doesn't throw in Safari version 10.1.2 (12603.3.8) :)

> How about doing a localStorage.clear() inside a before() call, like in here?

Yeah, that would be needed to remove anything from localStorage before doing the tests :) Shouldn't be necessary to clear the localStorage after [these tests](https://github.com/localForage/localForage/blob/d58233e73f6d5e4878b615a3576fcb4b6428e9b6/test/test.api.js#L163) as well? Just to keep the storage clean

So, if you think that it is ok I'm going to make the changes and rebuild the dist files :)

thgreasi:
> Shouldn't be necessary to clear the localStorage after these tests as well?

I have no strong opinion on that. Just adding a `before`, preparing the env for your tests, should be simpler than checking all other test cases for uncleared records.
I'm fine with this 👍 , just want an extra approval from @tofumatt as well, in case that I missed something.

jesusbotella:
@thgreasi Yeah, sure, there is not an only way to do it :) The only thing is that in the main tests, when they are all passed, there are items that are not removed.

I have just made the changes, and build dist files. :)

thgreasi:
Hey @jesusbotella, can you rebase over the current master and squash the commits of this PR a bit?

jesusbotella:
Yeah, sure. Do you want me to recreate dist files after rebasing and squashing commits?

thgreasi:
@jesusbotella yes, please! A separate dist commit would be ideal (in my mind) after squashing and rebasing the rest of the commits. That's not mandatory though but let's keep the commit history clean and short.

jesusbotella:
Done! I hope it is OK.

I have squashed all commits into one except the build ones, and then I made another one containing only the dist changes. :)

thgreasi:
:+1:

thgreasi:
Seems to work great even in IE9. Couldn't test Safari though.

jesusbotella:
I have just tested in Safari 10.1 (12603.1.30.0.34) and works as expected.

thgreasi:
@jesusbotella I know it's a bit late to ask this, but would this work equal well if we moved the above code in the [isLocalStorageValid method](https://github.com/localForage/localForage/blob/master/src/utils/isLocalStorageValid.js)?

jesusbotella:
Yes, I think so!

It will be checked before using localStorage, so I think it will work as expected. Let me test it, and modify the PR to make the change.

AVVS:
Would love to see this sooner than later! 👍

jesusbotella:
@thgreasi I have just changed the code to `isLocalStorageValid.js`, and it works like a charm. But I have one problem now.

I see that `isWebSQLValid.js` only checks if the `openDatabase` function exists, which returns true in our faulty drivers' test.
But as I changed the localStorage check to `isLocalStorageValid.js`, the localStorage driver is not valid, so the code doesn't fire the `setDriver` function success callback in the callback test, failing inevitably.

What can I do to pass the `callback` tests? If I set the error callback the localStorage test works, but the WebSQL test fails. Do you want me to split the test?

Another improvement that can be made in the promise test is to change that `.then(null, function() {})` to `.catch(function(){})`

thgreasi:
@jesusbotella oh, I didn't see that coming...
Let's keep this PR as is and merge it during the weekend.
The two approaches have both pros and cons and I'm not sure which I prefer the most.
* Moving this to _supports will result localstorage to be reported as not available in the supports() method as well but adds two store operations while localforage gets loaded and isn't consistent with what the rest drivers do. Also there is a change that the user gets an alert for allowing the site to store data, by the time the localstorage support checks run, even if they don't end up using localstorage at all.
* Keeping it inside the _initStorage is more consistent, since that's the place that the rest of the drivers do their first store operation.

CC @tofumatt for a review.

jesusbotella:
Ok! No worries :)

Yes, both approaches have pros and cons as you said, but I agree it wouldn't be right to show an alert in the user browser if they don't use localStorage at the end.

PR #748, opened 2017-10-03
Add missing require module name in build script

jesusbotella:
I made this pull request to add the `grunt` module name in build script string in the package.json. It makes the build script run properly as it was previously failing because of the missing name.

I use this package.json scripts a lot, as they are pretty handy when you try to install as less global packages as possible.

Thank you!

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
     "url": "git://github.com/localForage/localForage.git"
   },
   "scripts": {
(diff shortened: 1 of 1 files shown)


thgreasi:
Nice catch 👍

jesusbotella:
I thought it would be nice to add it, as it bothered me when I built the module dist files. :)
Thank you for merging it.

PR #749, opened 2017-10-04
Implement template strings for string interpolation; #740

### LINKED ISSUE #740: Use template strings for string interpolation
We should start using ES6's much nicer string interpolation instead of the string concatenation we currently do. Bonus points: we can use stuff like `oneLine` from the [common-tags](https://www.npmjs.com/package/common-tags) package like we do on the Add-ons repos.

tricinel:
Resolves #740


src/drivers/indexeddb.js
```diff
@@ -209,9 +208,9 @@ function _isUpgradeNeeded(dbInfo, defaultVersion) {
         // If the version is not the default one
         // then warn for impossible downgrade.
         if (dbInfo.version !== defaultVersion) {
-            console.warn('The database "' + dbInfo.name + '"' +
-                ' can\'t be downgraded from version ' + dbInfo.db.version +
-                ' to version ' + dbInfo.version + '.');
+            console.warn(
+                `The database ${dbInfo.name} can't be downgraded from version ${dbInfo.db.version} to version ${dbInfo.version}.`
```
thgreasi: Missing double quotes here around `${dbInfo.name}`

src/drivers/indexeddb.js
```diff
@@ -165,10 +165,9 @@ function _getConnection(dbInfo, upgradeNeeded) {
                     }
                 } catch (ex) {
                     if (ex.name === 'ConstraintError') {
-                        console.warn('The database "' + dbInfo.name + '"' +
-                            ' has been upgraded from version ' + e.oldVersion +
-                            ' to version ' + e.newVersion +
-                            ', but the storage "' + dbInfo.storeName + '" already exists.');
+                        console.warn(
+                            `The database ${dbInfo.name} has been upgraded from version ${e.oldVersion} to version ${e.newVersion}, but the storage ${dbInfo.storeName} already exists.`
```
thgreasi: Missing double quotes here around `${dbInfo.name}` & `${dbInfo.storeName}`

tricinel:
Ah, cool. I can look into common-tags. I actually never worked with it before.

src/drivers/indexeddb.js
tricinel: 👍

src/drivers/indexeddb.js
tricinel: 👍

tricinel:
Humm...so maybe I don't know how that oneLine things works...

thgreasi:
You probably need to use the [oneLine](https://www.npmjs.com/package/common-tags#oneline).

tricinel:
Yeah, that's the one I used...https://github.com/tricinel/localForage/blob/feature/es6-template-strings/src/drivers/indexeddb.js#L8

tricinel:
How keen are you on using that oneLine thing? I've reverted it in the mean time, and will need to rebase master...

tofumatt:
I like oneLine a fair bit but that’s probably because I’ve been using it a lot on other projects that have longer strings more often. Is it breaking something?

tricinel:
@tofumatt hmm, for some reason it was breaking the build...The tests weren't passing so I just assumed I was doing it wrong.

thgreasi:
I can take a look at it tomorrow 👍

thgreasi:
Hey @tricinel, it ends up that common tags increase the "nopromise" bundle size by 50% which sounds bad.
I tried something simpler... just running a regex func everywhere we use string templates.
See: https://github.com/thgreasi/localForage/tree/tricinel-feature/es6-template-strings

If you can squash your commits we can merge this during the next week 👍

tricinel:
Hey @thgreasi, thanks for taking a look. Quick question about squashing (forgive me, but this is my first time doing it). My workflow was:

- git rebase -i HEAD~3
- squashed the two oneLine commits into the previous one
- had to do git push origin +feature/es6-template-strings

Of course, now my branch is behind master, and rebase will show merge conflicts for the two dist/.min files. So...is it better to just start over? Close this one, start a new branch that's up to date with master, cherry pick the right commits from the old (this) feature branch and then submit a PR like that? Sounds like the cleanest option at the moment, but other ideas are welcome.

thgreasi:
@tricinel yea, rebasing with dist files included can be a pain... I tend to add the dist files as a seperate commit and drop them before rebasing & then generating & adding back after the rebase.
You can try squash everything and during the rebase pick the master's dist files as is and re-generate them later (that might be easier with an external non-console tool).
An alternative way to remove the dist file changes would be to pick the previous ones (from the old master) and them as a commit on top of yours and then squash your commits again.
I can do the rebase for you before merging If you pull out the dist files before merging.

tricinel:
@thgreasi, alright that should be it I guess. Thanks for the help!

tofumatt:
This was closed by #756 but we didn't close it, whoops! 😅

PR #756, opened 2017-10-19
refactor: use template strings

### LINKED ISSUE #740: Use template strings for string interpolation

thgreasi:
Merges #749 
Resolves #740 

@tofumatt how do you feel about this?


dist/localforage.js
```diff
@@ -1521,7 +1525,7 @@ function _initStorage$1(options) {
 
         // Create our key/value table if it doesn't exist.
         dbInfo.db.transaction(function (t) {
-            t.executeSql('CREATE TABLE IF NOT EXISTS ' + dbInfo.storeName + ' (id INTEGER PRIMARY KEY, key unique, value)', [], function () {
+            t.executeSql(oneLine('CREATE TABLE IF NOT EXISTS ' + dbInfo.storeName + '\n                     (id INTEGER PRIMARY KEY, key unique, value)'), [], function () {
```
thgreasi: We might want to avoid this call on every sqlite statement and keep it a single line to avoid any possible performance hit by running a regex.

thgreasi:
@tofumatt I totally agree with this, I also prefer longer lines than slower code.
@tricinel  tried oneLine at a274cf4 and some tests started to fail (!?) and the nopromise dist file increased by 50% size, since oneLine was included in the bundle and wasn't evaluated at compile time. Are you using any magical babel transformer [like this](https://github.com/WebEngage/babel-plugin-template-string-transform]?

tofumatt:
I’ll have to check on that but I think we have some fancy babel stuff around it. I won’t be able to get to it until probably the weekend though.

tricinel:
Hm, yes, I know. The tests were failing, but I couldn't figure out why...That's why I reverted it. The fact that it was dumping 50% more code in nopromise - I hadn't even checked, but that's equally bad.

Anything I can do to help?

PR #761, opened 2017-11-28
fix(indexeddb): cast null to undefined before storing value

### LINKED ISSUE #760: indexedDB on Edge doesn't support putting null values
Trying to call `localforage.setItem("foo", null)` rejects with error `Invalid argument`. According to [this bug](https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/106663/), indexeddb on edge currently does not support null values.
The code does contain a workaround as far as i can see (although it's commented as a workaround for IE10): https://github.com/localForage/localForage/blob/b55bb06801b2d2dd65b64b001db2d31bf44be138/src/drivers/indexeddb.js#L587
However the value is cleaned _after_ the call that fails (`store.put(value, key)`, line 581). I do believe that it should be cleaned up before that, since the value is only returned to the user after that and no longer used anywhere else in the code.
For now, I will just note this and set it to `undefined` instead of `null`.

tran2:
to support Microsoft Edge
fixes #760


thgreasi:
I think that this should be in v1.6, since master already contains dropInstance(). Sorry about that, will bump the version during the weekend.

tofumatt:
Oh shoot, you’re right. I already released 1.5.4 but I’ll see about releasing a 1.5.5 that rolls it back I guess. And then releasing a 1.6.

- Matthew Riley MacPherson
tofumatt.com

PR #767, opened 2017-12-09
Export typings

### LINKED ISSUE #762: 1.5.4 typings broke signature
Hi,

In 1.5.4 `defineDriver` started expecting `dropInstance` method in driver configuration object. Preferably signature changes are released as major versions (2.0.0 ?). Or maybe `dropInstance` could be made optional.

Consider also adding `dropInstance` documentation to this section:
https://github.com/localForage/localForage/blob/master/docs/api.md#driver-api

Thanks

### LINKED ISSUE #764: Why are not most interfaces exported?
The "localforage.d.ts" has some interfaces like LocalForage, LocalForageOptions, LocalForageDbMethods.

But any interfaces is not exported. Exported is only LocalForage's variable.
At least, I think `LocalForageOptions` should be exported.
Otherwise, even if the developer uses LocalForage with TypeScript, we can not know what option LocalForageOptions has.

Why are not most interfaces exported in "localforage.d.ts" ?

thgreasi:
Resolves #762
Resolves #764


thgreasi:
@ponday-dev, @chetmurphy , @alan-agius4, @michaelgeorgeattard Can I get some feedback on this?

chetmurphy:
The pull request works for my case. LocalForage type is available and does not generate any type error.

Thanks.

thgreasi:
These changes seems to break the typings of the various localforage plugins (setItems etc).
Keeping this open and adding it to the v2.0 milestone.

PR #776, opened 2018-01-05
Add jsDelivr hits badge

LukasDrgon:
Hey, I noticed that localForge gets 45k hits per month on jsDelivr, so I thought you might like this badge.


thgreasi:
Cool thanks 👍

PR #786, opened 2018-01-21
fix(indexedDb): close connections to fix dropInstance timeouts

### LINKED ISSUE #779: dropInstance test occasionally timeout
On a fresh clone created like this:

```
git clone https://github.com/localForage/localForage.git
cd localForage
npm install
bower install
```

Tests hang at this line:
```
npm test
[...]
Running "mocha:unit" (mocha) task
Testing: http://localhost:9999/test/test.main.html
```

Same thing happens with `grunt test`.

Versions:
Node: v8.6.0
Npm: 5.6.0
Bower: 1.8.2
macOS: 10.12.6 (16G1114)

thgreasi:
Resolves: #779 
See:  #708

### DIFF
--- a/src/drivers/indexeddb.js
+++ b/src/drivers/indexeddb.js
@@ -10,14 +10,14 @@ import getCallback from '../utils/getCallback';
 // Some code originally from async_storage.js in
 // [Gaia](https://github.com/mozilla-b2g/gaia).
 
-var DETECT_BLOB_SUPPORT_STORE = 'local-forage-detect-blob-support';
-var supportsBlobs;
-var dbContexts;
-var toString = Object.prototype.toString;
+const DETECT_BLOB_SUPPORT_STORE = 'local-forage-detect-blob-support';
+let supportsBlobs;
+const dbContexts = {};
+const toString = Object.prototype.toString;
 
 // Transaction Modes
-var READ_ONLY = 'readonly';
-var READ_WRITE = 'readwrite';
+const READ_ONLY = 'readonly';
+const READ_WRITE = 'readwrite';
 
 // Transform a binary string to an array buffer, because otherwise
 // weird stuff happens when you try to work with the binary string directly.
@@ -143,6 +143,8 @@ function _rejectReadiness(dbInfo, err) {
 
 function _getConnection(dbInfo, upgradeNeeded) {
     return new Promise(function(resolve, reject) {
+        dbContexts[dbInfo.name] = dbContexts[dbInfo.name] || createDbContext();
+
         if (dbInfo.db) {
             if (upgradeNeeded) {
                 _deferReadiness(dbInfo);
@@ -318,18 +320,21 @@ function _tryReconnect(dbInfo) {
             forage._dbInfo.db = null;
         }
     }
+    dbInfo.db = null;
 
     return _getOriginalConnection(dbInfo)
         .then(function(db) {
-            for (var j = 0; j < forages.length; j++) {
-                forages[j]._dbInfo.db = db;
-            }
             dbInfo.db = db;
-        })
-        .then(function() {
             if (_isUpgradeNeeded(dbInfo)) {
                 // Reopen the database for upgrading.
                 return _getUpgradedConnection(dbInfo);
+            }
+            return db;
+        })
+        .then(function(db) {
+            dbInfo.db = dbContext.db = db;
+            for (var j = 0; j < forages.length; j++) {
+                forages[j]._dbInfo.db = db;
             }
         })
         .catch(function(err) {
@@ -385,6 +390,19 @@ function createTransaction(dbInfo, mode, callback, retries) {
     }
 }
 
+function createDbContext() {
+    return {
+        // Running localForages sharing a database.
+        forages: [],
+        // Shared database.
+        db: null,
+        // Database readiness (promise).
+        dbReady: null,
+        // Deferred operations on the database.
+        deferredOperations: []
+    };
+}
+
 // Open the IndexedDB database (automatically creates one if one didn't
 // previously exist), using any options set in the config.
 function _initStorage(options) {
@@ -399,26 +417,12 @@ function _initStorage(options) {
         }
     }
 
-    // Initialize a singleton container for all running localForages.
-    if (!dbContexts) {
-        dbContexts = {};
-    }
-
     // Get the current context of the database;
     var dbContext = dbContexts[dbInfo.name];
 
     // ...or create a new context.
     if (!dbContext) {
-        dbContext = {
-            // Running localForages sharing a database.
-            forages: [],
-            // Shared database.
-            db: null,
-            // Database readiness (promise).
-            dbReady: null,
-            // Deferred operations on the database.
-            deferredOperations: []
-        };
+        dbContext = createDbContext();
         // Register the new context in the global container.
         dbContexts[dbInfo.name] = dbContext;
     }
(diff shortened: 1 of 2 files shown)


thgreasi:
This needed a huge effort, regardless the low +-diff, but I wanted to get this fixed before my family gets bigger 👶 .
With this, we should be good to release v1.6 😅 and start planning v2.0 (huge excitement)!

@jlchereau the excellent tests that you provided pass with this PR. Tested both on FF & Chrome.
Would love to get any additional feedback on this.

src/drivers/indexeddb.js
```diff
@@ -318,19 +320,22 @@ function _tryReconnect(dbInfo) {
...
-        })
-        .then(function() {
             if (_isUpgradeNeeded(dbInfo)) {
                 // Reopen the database for upgrading.
                 return _getUpgradedConnection(dbInfo);
             }
+            return db;
+        })
+        .then(function(db) {
+            dbInfo.db = dbContext.db = db;
```
thgreasi: Keep the latest db reference, in case the db needed upgrade.

src/drivers/indexeddb.js
```diff
@@ -399,26 +417,12 @@ function _initStorage(options) {
         }
     }
 
-    // Initialize a singleton container for all running localForages.
-    if (!dbContexts) {
-        dbContexts = {};
```
thgreasi: Moved this to the top of the file.

src/drivers/indexeddb.js
```diff
@@ -947,36 +951,57 @@ function dropInstance(options, callback) {
...
-        var dbPromise =
-            options.name === currentConfig.name && self._dbInfo.db
-                ? Promise.resolve(self._dbInfo.db)
-                : _getOriginalConnection(options);
+        const isCurrentDb =
+            options.name === currentConfig.name && self._dbInfo.db;
+
+        const dbPromise = isCurrentDb
+            ? Promise.resolve(self._dbInfo.db)
+            : _getOriginalConnection(options).then(function(db) {
```
tofumatt: Any reason we're using `function(db) {` here over `(db) => {`? I'm just used to seeing the more ES6-y type functions these days so I'd prefer we use them in the future if we can.

src/drivers/indexeddb.js
```diff
@@ -947,36 +951,57 @@ function dropInstance(options, callback) {
...
+                  const dbContext = dbContexts[options.name];
+                  const forages = dbContext.forages;
+                  dbContext.db = db;
+                  for (var i = 0; i < forages.length; i++) {
```
tofumatt: Seeing the `var` and also the regular `for` loop here doesn't seem very ES6-like either, can we do it a bit more functionally?

src/drivers/indexeddb.js
```diff
@@ -947,36 +951,57 @@ function dropInstance(options, callback) {
...
-                        forage._dbInfo.db = null;
-                    }
+                    forage._dbInfo.db = null;
                 }
 
                 var dropDBPromise = new Promise(function(resolve, reject) {
                     var req = idb.deleteDatabase(options.name);
 
-                    req.onerror = req.onblocked = reject;
+                    req.onerror = req.onblocked = e => {
```
tofumatt: I'd like an eslint rule on this in general, but I think especially here as we already have a few `=` on the line, it'd be nice to do `(e) => {` instead of `e => {` because I find the parenthesis help indicate a function a lot better.

Also, could we use `err` or `error` here instead? I know `e` usually means error but I dislike single-lettle variables 😄

src/drivers/indexeddb.js
```diff
@@ -994,42 +1019,49 @@ function dropInstance(options, callback) {
...
 
-                for (var i = 0; i < forages.length; i++) {
-                    var forage = forages[i];
-                    if (forage._dbInfo.db) {
-                        forage._dbInfo.db.close();
-                        forage._dbInfo.db = null;
-                        forage._dbInfo.version = newVersion;
-                    }
+                db.close();
+                for (let i = 0; i < forages.length; i++) {
```
tofumatt: I don't think we should be using `let` inside `for` loops, [because JS is weird](https://stackoverflow.com/questions/16473350/let-keyword-in-the-for-loop) though the bugs there are kind of fixed. But we aren't re-assigning `i` at any rate for each loop, so `const` still makes more sense.

src/drivers/indexeddb.js
```diff
@@ -994,42 +1019,49 @@ function dropInstance(options, callback) {
...
+                    const forage = forages[i];
+                    forage._dbInfo.db = null;
+                    forage._dbInfo.version = newVersion;
                 }
 
-                var dropObjectPromise = new Promise(function(resolve, reject) {
-                    var req = idb.open(options.name, newVersion);
+                const dropObjectPromise = new Promise(function(
```
tofumatt: As mentioned before, it'd be cool if we used fat arrow (`(resolve, reject) => {}`) functions where we can going forward.

src/drivers/indexeddb.js
```diff
@@ -994,42 +1019,49 @@ function dropInstance(options, callback) {
...
+                    resolve,
+                    reject
+                ) {
+                    const req = idb.open(options.name, newVersion);
 
-                    req.onerror = reject;
+                    req.onerror = e => {
```
tofumatt: As mentioned too, I'd like to see us wrap things like this `e` in parens and use no single-letter variables, so it'd be cool if this could be `req.onerror = (err) => {`

src/drivers/indexeddb.js
```diff
@@ -994,42 +1019,49 @@ function dropInstance(options, callback) {
...
                         resolve(db);
                     };
                 });
 
                 return dropObjectPromise
                     .then(function(db) {
-                        for (var j = 0; j < forages.length; j++) {
-                            var forage = forages[j];
+                        dbContext.db = db;
+                        for (let j = 0; j < forages.length; j++) {
```
tofumatt: Should be `const` too, if we stay with the loop.

src/drivers/indexeddb.js
```diff
@@ -318,19 +320,22 @@ function _tryReconnect(dbInfo) {
...
-        .then(function() {
             if (_isUpgradeNeeded(dbInfo)) {
                 // Reopen the database for upgrading.
                 return _getUpgradedConnection(dbInfo);
             }
+            return db;
+        })
+        .then(function(db) {
+            dbInfo.db = dbContext.db = db;
+            for (var j = 0; j < forages.length; j++) {
```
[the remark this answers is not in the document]
thgreasi: Works be really happy to do such a thing.
What do you have in mind?
Last time I checked `.forEach` wasn't getting polyfilled in this repo and `for of` had an other issue (will try again later).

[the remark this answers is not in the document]
thgreasi: Perhaps in v2.0 we could have one build with polyfills (not just Promise) and one without. We should probably also default to the non polyfilled version. I've read about people being concerned of the main dist file size without knowing that we have a nopolyfills variant.

src/drivers/indexeddb.js
thgreasi: Oh my, browser vendors are weird

thgreasi: But, [const  isn't allowed in for loops](https://babeljs.io/repl/#?babili=false&browsers=IE%208&build=&builtIns=usage&code_lz=GYVwdgxgLglg9mABMAFDAlIg3gKEYiBAZzgBsBTAOlLgHM10BuHAXxx0LCKkQC9EAvIgDaAcgCGogDSiARtNERRAXWbA4AJ0QpO3RDEGIADI32IAPH2rkwtKAAtTMANTPMuAJBFyUACowAW3I4ECgUFEwBAD5kBiZWHCA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=true&lineWrap=true&presets=es2015-loose&prettier=false&targets=&version=6.26.0&envVersion=1.6.2) and `let` will be transpiled anyway.
What do you suggest?

src/drivers/indexeddb.js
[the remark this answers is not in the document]
thgreasi: @tofumatt Using babel es2015-loose [`for of` & `.forEach()`](https://babeljs.io/repl/#?babili=false&browsers=IE%208&build=&builtIns=usage&code_lz=PTAEEsFsAcHsCcAuoBEAjAhmgpgGwLRy4CeAZuLrigNwBQtpArgHYDGi4szopAFOAEpQAb1qhQrLgGdYubADpcsAOb8BdAL71JzKcgBeoALygA2gHIM5gDTm0N863MBdOqQSheOvaAAeoWFJQfSFRAEg-X3VaLQYPXjlkcGNQAAZqCFAAHmDFbGZlRAALDPAAajLQ2jCpbEQAFShsWEZEXl4hIwA-HjVo2P15SAxodqjjHsiBaKA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=true&lineWrap=true&presets=es2015-loose&prettier=false&targets=&version=6.26.0&envVersion=1.6.2) works on IE9+,  as long as you only use arrays, otherwise the reference to `Symbol` in the transpiled code will error when it gets evaluated.
My concern (which holded me back all this time) include:
* is the extra file size increase of the transpiled `for...of` really worth it?
* should we start using untranspiled es5 Array.* methods?

PS: Typescript (because of the type inference) [emits far less & cleaner code for `for...of`](https://www.typescriptlang.org/play/index.html). Should we consider it at some point? Especially editing this driver would feel safer.

src/drivers/indexeddb.js
thgreasi: Changed that, added an eslint rule, but in the end prettier didn't like it and I couldn't find a way to change its mind :(

src/drivers/indexeddb.js
thgreasi: TBH, I was just doing this for consistency in the code style :) Will be more than happy to change these.

src/drivers/indexeddb.js
thgreasi: See earlier comment
> [const  isn't allowed in for loops](https://babeljs.io/repl/#?babili=false&browsers=IE%208&build=&builtIns=usage&code_lz=GYVwdgxgLglg9mABMAFDAlIg3gKEYiBAZzgBsBTAOlLgHM10BuHAXxx0LCKkQC9EAvIgDaAcgCGogDSiARtNERRAXWbA4AJ0QpO3RDEGIADI32IAPH2rkwtKAAtTMANTPMuAJBFyUACowAW3I4ECgUFEwBAD5kBiZWHCA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=true&lineWrap=true&presets=es2015-loose&prettier=false&targets=&version=6.26.0&envVersion=1.6.2)

thgreasi:
@tofumatt I just added even more tests for this (and my local git broke just before pushing... sad story).
Answering to some of your requests, I provided some details on why I was still using plain `for` loops.
Please give me a hint whether this needs further changes to merge. Feel free to push any commits, merge and publish v1.6 yourself in case I disappear for a while (👶 is almost here).

thgreasi:
Ping @tofumatt for re-review!

src/drivers/indexeddb.js
tofumatt: Oh shoot, I realised that, we should do `let`. My brain got confused because in `for... in/of` loops it makes sense to use `const` I think, which we do in the Mozilla addons JS. This is fine 😄

thgreasi:
@tofumatt Should I cut a release for v1.6 or do we want to add anything else before releasing?

thgreasi:
@tofumatt Should I cut a release for v1.6 or do we want to add anything else before releasing?

tofumatt:
Oh shoot, sorry I missed this.

I think we're all good, let's ship it! 🚢

Issue #658, opened 2017-02-02
The "version" configuration option is not clearly documented.


lddubeau:
The documentation for the `version` configuration option currently states:

> version
> The version of your database. May be used for upgrades in the future; currently unused.

The problem is that this does not clarify whether this value is something that the client code may use. "May be used for upgrades in the future" does not clarify because it could mean that eventually localForage may have facilities to detect and notify client code when localForage finds an old version of the database that the client code should upgrade. Actually #117 shows that such an idea has been floated in the past. A user following this interpretation may decide to fill the field with whatever version number suits them with the idea that they'll (somehow) handle the version changes themselves for now until localForage provides some sort of formal support.

However, upon reading the code, I see localForage has its own `onupgradeneeded` handler where it may do perform actions internal to localForage. So it seems to me that this number is something that should be used *only* if localForage's documentation says to use another value. It should be reserved for upgrades *of localForage's internal structures*, and not for upgrades to the structures accessible to the client code. I would not call it "unused". It is used to set the database version as understood by IndexedDB (and WebSQL, too I assume). If it were unused, I'd expect that setting it to 2, 30, or 1243 would not matter, but people setting it to such values may run into trouble later when localForage's documentation eventually states 

> to take advantage of localForage's new feature X, your database must be upgraded to version 2.

The poor dev who has already set it to 2 will find himself unable to upgrade, and those who are at higher values will get a warning about a downgrade and probably erratic behavior.

(As a aside, an additional thing that makes me question the value of this `version` option is that it seems the `localStorage` backend simply tosses the value aside. When I look into a `localStorage` instance populated by localforage, I cannot see the version value stored anywhere. Did I miss something?)

Even if my inferences are not correct above, I'd still say the role this option may play in the future is so nebulous right now that until its role is crystal-clear the documentation should not say that it is unused but instead tell users to not use it. (Or remove it from the options??)

paulpooch:
If you accidentally pass a semver style string to the version option.  All of local forage just silently fails.  Very confusing.

thgreasi:
I would expect that .ready() would reject in that case. Would you prefer if
it threw an error right away?

> --
*Thodoris Greasidis*

paulpooch:
It did not reject.  I think the promise just hangs and never resolves.  No
catch blocks detected any issue.

I mean it's not a huge deal just maybe clarify the docs a little.  Looks to
me like indexedDB just runs on int versioning... It took me an hour of
wasted time to figure that out.

[...]

Issue #117, opened 2014-03-20
Provide a DB version change API


tofumatt:
Re: #116 -- we don't currently handle upgrade events, even in drivers that have very good support for upgrades (i.e. IndexedDB). We don't really employ a schema, but maybe it would make sense to provide an API to allow people to do data migrations across versions of their DB?

Or is this out-of-scope for localForage, and something an app/framework should handle?

augbog:
Stumbling on this as I was wondering if localForage provided a way to help handle upgrades -- I personally would be very interested in this especially if we were able to provide a way for localStorage and WebSQL a way to migrate. I see packages like [angular-localforage-migrations](https://github.com/psalaets/angular-localforage-migrations) that wouldn't be as necessary if this were the case.

My original instinct that localForage versioning would somehow hook into IndexedDB but #658 showed that wasn't the case (thankfully there has been healthy discussion about this :D). Would like to see more thoughts around this though as I feel it would be a very useful feature.

tofumatt:
Unfortunately right now upgrades aren’t really handled well by localForage :-(

It’d be an awesome feature to add but we haven’t been able to get to it.

thgreasi:
It should be easy to make an plain js migration plugin based on what the angular one already does. I imagine having a main LF core (this repo) and several satellite plugins under the other. It would be easy to implement independently, evaluate & reiterate. It would also ease evaluating the implementation & tracking user acceptance (eg npm DLs).

On the other hand I'm thinking about a `__localforage__system__` instance/db/store for some time now, holding other things like the used instances.

augbog:
@thgreasi yeah I actually didn't mean to suggest that the migration plugins aren't good -- if that's the preferred route for people to incorporate migrations in their apps, I just feel that should be specified somewhere (essentially #116) 😛

PR #789, opened 2018-02-06
#658 add documentation for version configuration

augbog:
From the original issue #658 there was talk about how the `version` configuration actually does do specific things in localForage. Testing locally it does seem we do use it.

In WebSQL, it does seem we pass the configuration in but from there I could not see anything WebSQL does with it. In IndexedDB, it also uses the configuration version set and sets it as IndexedDB's version. If we detect that it was an upgrade (`_isUpgradeNeeded`), we do call a new connection, pass in there was an upgrade and `localForage` will define its own `onupgradeneeded` event which will attempt to recreate the object store. There also seems to be some logic where when a new store is created, it will also trigger this and increment IndexedDB's version by 1 on its own.

I'm not sure if the documentation I wrote clearly suggests all that I wrote so if someone has suggestions for improvements, I'm totally willing to make them! Hope this is accurate from what I can find.

Also FYI two things I didn't document but might be worth mentioning

1) There was a mention of semantic versioning in the PR but I believe it's accurate that we are doing a `typeof number` because nowhere did I find that IndexedDB supports semver so I just tried to be clear that it was a "schema version"

2) If an upgrade is detected, we call `db.createObjectStore(dbInfo.storeName)` which ends up potentially throwing `ConstraintError`. This is actually expected if the user did not clear the original storeName [because `createObjectStore` only can create if its a unique name](https://stackoverflow.com/questions/17770716/constraint-error-on-indexeddb-database-when-updating-version-and-nothing-else#answer-17772759) :/ not sure if this is something that needs to be addressed.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -367,7 +367,7 @@ Set and persist localForage options. This must be called *before* any other call
   </dd>
   <dt>version</dt>
   <dd>
-    The version of your database. May be used for upgrades in the future; currently unused.<br>
+    The schema version of your database. Used only in WebSQL and IndexedDB. In WebSQL, this simply sets the version, and in IndexedDB this may trigger an <code>onupgradeneeded</code> event if a version upgrade is detected. If a new store is detected, localForage will ask IndexedDB to increment the version itself to manually trigger the <code>onupgradeneeded</code> event. As of right now, upgrade events are not customizable, but may be in the future. For drivers that do not support configuration for versioning, this value simply gets thrown away.<br>
     Default: <code>1.0</code>
   </dd>
   <dt>description</dt>


augbog:
Yeah if there are thoughts on how to clean it up and/or more concise, I'm totally willing to make those changes! Also, if it makes sense, linking to the issue/discussion in the documentation is another option to help make it not seem as verbose.

augbog:
Hey just curious if this is sufficient to merge? I know people are busy so no worries if busy! Just wanted to check in :)

tofumatt:
Oh sorry, I think it just got missed after the review. I agree that maybe it's too technical but it's better than what we had before, so I'll merge it for now and we can always tidy it up later! 😄

augbog:
Thanks @tofumatt! :) Yeah if there's feedback on how to improve it, more than willing to help update! Also looking into #117 although best option might be to make a plugin as suggested.

moonlitknight:
I need to add indexeddb versioning to my app. I don't need localForage to do much more than expose the existing version so I can detect that an upgrade is needed. Is this possible?

thgreasi:
@moonlitknight atm localforage doesn't provide an API to interact with the DB versions and upgrades. It actually tries to conceal upgrading whenever needed, so that you don't need to handle it at all, and your get/setItem calls eventually fulfill, regardless of the store and its state.

PR #807, opened 2018-03-29
Avoid uncaught error in dropInstance onblocked handler

AdamStone:
[Context](https://github.com/localForage/localForage/issues/800)

I found I had to make the change to .prettierrc to avoid the precommit step replacing all the ' with "

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,4 +1,4 @@
 {
-  "single-quote": true,
+  "singleQuote": true,
   "tabWidth": 4
 }
--- a/src/drivers/indexeddb.js
+++ b/src/drivers/indexeddb.js
@@ -198,7 +198,16 @@ function _getConnection(dbInfo, upgradeNeeded) {
         };
 
         openreq.onsuccess = function() {
-            resolve(openreq.result);
+            var db = openreq.result;
+            db.onversionchange = function(e) {
+                // Triggered when the database is modified (e.g. adding an objectStore) or
+                // deleted (even when initiated by other sessions in different tabs).
+                // Closing the connection here prevents those operations from being blocked.
+                // If the database is accessed again later by this instance, the connection
+                // will be reopened or the database recreated as needed.
+                e.target.close();
+            };
+            resolve(db);
             _advanceReadiness(dbInfo);
         };
     });
@@ -984,12 +993,22 @@ function dropInstance(options, callback) {
                 const dropDBPromise = new Promise((resolve, reject) => {
                     var req = idb.deleteDatabase(options.name);
 
-                    req.onerror = req.onblocked = err => {
+                    req.onerror = () => {
                         const db = req.result;
                         if (db) {
                             db.close();
                         }
-                        reject(err);
+                        reject(req.error);
+                    };
+
+                    req.onblocked = () => {
+                        // Closing all open connections in onversionchange handler should prevent this situation, but if
+                        // we do get here, it just means the request remains pending - eventually it will succeed or error
+                        console.warn(
+                            'dropInstance blocked for database "' +
+                                options.name +
+                                '" until all open connections are closed'
+                        );
                     };
 
                     req.onsuccess = () => {


tofumatt:
Agreed, it would be good to have a test case for this… could we add one for FF and just run it in the browser for now to make sure it works? I'm gonna try to spend Sunday getting Karma or something similar working, for what it's worth 😄

AdamStone:
@thgreasi @tofumatt I manually tested this a bit and it seems it doesn't work the way I expected. I had taken the blocked event to mean that the request would not proceed, but in fact it remains pending until the other tab's connection is closed, at which point the `onsuccess` handler fires. So the database is eventually dropped, but in the meantime it hangs. [This discussion](https://stackoverflow.com/questions/40032008/how-should-an-app-react-when-indexeddb-is-blocked) seems useful to figure out a better approach here.

AdamStone:
Case #3 in the link above was the behavior I was looking for, but unfortunately the given example for how to accomplish it does not actually work. In the case of a deleteDatabase request, `request.transaction` is undefined so `request.transaction.abort()` doesn't work. I actually don't see any way to abort a deleteDatabase request once created.

AdamStone:
My next thought was to try to check whether the database was in use before trying to drop it. So far I couldn't find a way to do this either. Maybe I'm missing something, but it looks like the only tool the API offers to deal with this is the `onversionchange` event (on delete the version appears to change to null). That would enable one tab to be notified that another is deleting the database so it can close its own connection to it. But if that tab still needs the database, it would seem the only option is to copy the whole thing to a new instance before closing the connection. And if the version change event is ignored, subsequent writes silently fail in the sense that they'll be wiped out as soon as the tab is closed.

thgreasi:
Should we close this for now?

AdamStone:
I'd like to fix it somehow, as the current behavior throws an error and the promise hangs. I don't see a clear solution though. It seems wrong to reject on blocked if eventually the request will finish, and it seems wrong to resolve on blocked since the request is still pending. If neither is done, the promise will hang until one of the connections is closed.

AdamStone:
@thgreasi I updated the PR based on what I described [here](https://github.com/localForage/localForage/issues/821), this seems to resolve that issue and prevents the delete request from being blocked as well. Just in case, I left a warning in the onblocked handler. It still leaves the possibility of having a database deleted by one session out from under another that's still using it, as I still couldn't find a way to abort a delete request once made, but that seems better than having the session continue to write to a database that will be deleted later after the connection closes. 

Is there an existing test suite for indexedDB where I could try to add tests for this?

thgreasi:
You can use `npm run serve` and at some point a url like https:0.0.0.0/test/main.html will show up. You can use that on your browser.

ReasonableDeveloper:
I'm running into the same problem trying to drop a database from my main Client with a service-worker also interacting with the localforage. Any chances of getting this into master @tofumatt @thgreasi?

koriwi:
same here

AdamStone:
For what it's worth, I'm no longer confident this is a good change, at least in isolation. Although it seems to solve the immediate problem of the hanging request, I continued to run into errors when version changes affected a DB that was in use in multiple tabs (maybe the version warning is not really innocuous). In my own application I ended up just using a different design to avoid these kinds of collisions (creating a separate DB for each namespace and using only the default object store, so the version remains stable when multiple tabs use that namespace), because I couldn't get it working reliably otherwise.

ahayes91:
> For what it's worth, I'm no longer confident this is a good change, at least in isolation. Although it seems to solve the immediate problem of the hanging request, I continued to run into errors when version changes affected a DB that was in use in multiple tabs (maybe the version warning is not really innocuous). In my own application I ended up just using a different design to avoid these kinds of collisions (creating a separate DB for each namespace and using only the default object store, so the version remains stable when multiple tabs use that namespace, and just using `store.clear()` instead of `dropInstance`), because I couldn't get it working reliably otherwise.

Hey @AdamStone you mentioned you avoid dropInstance when using localForage across multiple tabs - this could be a n00b question, but do you know if there are any performance implications if you leave orphaned IndexedDB databases in the browser? I'm having trouble with dropInstance and thinking I might just put the effort into making sure the databases are cleared on logout / browser tab close instead, but feels a bit untidy to leave those empty databases lying around...

AdamStone:
I'm not sure if there's any performance impact from total number of databases, but I think as a matter of principle I'd try to avoid designs that would create lots of stale databases that keep accumulating unbounded (like making a new one each day). I'd look for a way to structure the data to enable mostly reusing the same databases over time, and only occasionally needing to add new ones.

Hivemind9000:
It seems that this conversation has "hung" just like the driver does when blocked. ;-)

Can't this PR be approved/merged (with the small change below) so at least we developer/users can take control when the problem occurs? As it currently stands, we can't do anything about it except avoid using dropInstance altogether.

If the req.onblocked section of @AdamStone's code could also throw an error, then I could catch it and take action. I know it's not ideal, but it's better than the current black hole of no response. I'm not certain of how to close and reopen the IndexedDB connection via localforage's API though. If the _tryReconnect method could be exposed somehow (e.g. localforage.reconnect()), then wouldn't this free up the lock so dropInstance could proceed?

tofumatt (approved):
That's a fair point and we can totally merge this for now as-is. I don't think it'll cause any issues. I'll merge it now and cut a release later today (it will be a minor version update because of the change).

I'll update this PR when the release is out.

tofumatt:
I just published version `1.10.0` of localForage, which includes this fix. Please let me know if that's working for you to catch these errors for now! 🙂

Hivemind9000:
Thanks for the quick response @tofumatt.

I can confirm that I am now able to use dropInstance (dropping a database rather than just an object store within) and it completes perfectly without error or delay.

I also (temporarily) reverted my code back to using a single database with multiple object stores, and the multi-tab conflicts (onblocked) I was getting when I created a new object store (while the other tab had a different object store open) have now been resolved too.

As a matter of precaution I think I'll stick to the one database/one object store just to be on the safe side.

Thanks again! 🙂

Issue #824, opened 2018-05-31
IDBTransaction randomly hangs for concurrent operations in Safari/WebKit


ianweng:
Running below test cases in Safari, some of they will timeout. You can try to change the timeout to 0 (infinite long), then it will hang forever.

Environments:
mac OS 10.13.4 (17E202)
iOS 11.3.1 (15E302)

```
<!doctype html>
<html lang="en">
  <head>
    <meta charset="utf-8">
    <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/mocha/5.2.0/mocha.css">
  </head>
  <body>
    <div id="mocha"></div>
    <script src="https://cdnjs.cloudflare.com/ajax/libs/localforage/1.7.1/localforage.js"></script>
    <script src="https://cdnjs.cloudflare.com/ajax/libs/mocha/5.2.0/mocha.js"></script>
    <script>
      mocha.setup({ ui: 'bdd', timeout: 30000 });
      window.onload = function() { mocha.run(); };
      test('getItem', (i) => i.getItem('key')); // passing
      test('setItem', (i) => i.setItem('key', 'value')); // failing, randomly hangs
      test('clear', (i) => i.clear()); // failing, randomly hangs
      function test(name, op) {
        describe(name, function() {
          for (let i = 0; i < 500; i++) {
            it('test' + i, function(done) {
              const instances = [
                localforage.createInstance({ name: 'db1', storeName: 't1' }),
                localforage.createInstance({ name: 'db1', storeName: 't2' }),
                localforage.createInstance({ name: 'db1', storeName: 't3' }),
                localforage.createInstance({ name: 'db2', storeName: 't1' }),
                localforage.createInstance({ name: 'db2', storeName: 't2' }),
                localforage.createInstance({ name: 'db2', storeName: 't3' }),
                localforage.createInstance({ name: 'db3', storeName: 't1' }),
                localforage.createInstance({ name: 'db3', storeName: 't2' }),
                localforage.createInstance({ name: 'db3', storeName: 't3' }),
              ];

              Promise
                .all(instances.map(op))
                .then(() => done(), () => done());
            });
          }
        });
      }
    </script>
  </body>
</html>
```

ianweng:
From what I found
1) none of onabort/onerror/oncomplete of the transaction (IDBTransaction) created by localforage.clear() are fired which should be a WebKit issue but I believe localForage should play around with it.
[...]

PR #826, opened 2018-06-07
Workaround webkit indexeddb hanging transaction

ianweng:
link #824


src/drivers/indexeddb.js
```diff
@@ -775,6 +794,7 @@ function clear(callback) {
             })
             .catch(reject);
     });
+    promise.then(unref, unref);
```
Shingyx: The current solution will retain the transactions in memory forever and never free them. This is because when `promise.then(unref, unref);` is executed, `unref` is still `undefined` since the callback hasn't executed yet.

To get around this, you could move the `promise.then(unref, unref);` inside the callback, or wrap the unref calls inside another function, e.g. something like:

```
promise.then(
  () => unref && unref(),
  () => unref && unref()
);
```

src/drivers/indexeddb.js
```diff
@@ -345,6 +345,38 @@ function _tryReconnect(dbInfo) {
         });
 }
 
+// Safari could garbage collect transaction before oncomplete/onerror/onabord being dispatched
```
Shingyx: Typo here. "onabord" should be "onabort".

src/drivers/indexeddb.js
```diff
@@ -345,6 +345,38 @@ function _tryReconnect(dbInfo) {
...
+function createReadWriteTransactionHelper() {
+    var unref = undefined;
+    return {
+        create: function(dbInfo, callback, retries) {
+            createTransaction(
+                dbInfo,
+                READ_WRITE,
+                function(err, transaction) {
+                    var id = _refTransactionId++;
+                    _refTransaction[id] = transaction;
```
Shingyx: Technically very unlikely to happen, but it would be safer to `% Number.MAX_SAFE_INTEGER` the `_refTransactionId` for applications which run for a long time.

src/drivers/indexeddb.js
```diff
@@ -345,6 +345,38 @@ function _tryReconnect(dbInfo) {
...
+                        delete _refTransaction[id];
+                    };
+                    callback(err, transaction);
+                },
+                retries
+            );
+        },
+        done: function(promise) {
+            var lazyUnref = function() {
+                unref && unref();
```
Shingyx: Might be better to have the following since jshint doesn't like `unref && unref()` (jshint is enabled on https://github.com/localForage/localForage-removeItems).

```
if (unref) {
  unref();
}
```

PR #830, opened 2018-06-15
fix(websql): fix dropped detection 

### LINKED ISSUE #829: Google Closure Compiler fails because of a missing name variable
Hey all,

I'm trying to compress my project via Google Closure Compiler via the advanced mode but it's sadly falling over because of this line here:

https://github.com/localForage/localForage/blob/e4b4c9427ef3e168c1f04fb29bb9e3468f4828ff/dist/localforage.js#L1712

The error is:
```
ERROR - variable name is undeclared
            t.executeSql('SELECT name FROM sqlite_master ' + "WHERE type='table' AND name = ?", [name], function (t, results) {
```
I'm inclined to agree with the compiler here, what is the name variable meant to be pointing to? I was going to make a PR but I can't really make sense of this function.

If anyone could help that would be really appreciated.

Many thanks,
Ash

thgreasi:
Resolves #829


ashpr:
Thank you so much @thgreasi . Can confirm this fixes it. No complaints from GCC!

PR #874, opened 2019-02-08
Switch use of IndexedDB openCursor to openKeyCursor for fetching keys

MeMark2:
First off, first time contributing to this project. Let me know if I'm doing anything wrong.

A little bit of background:
I'm running into a problem with some of my production IndexedDB databases which seems to be related to accessing the whole db with `openCursor`. I haven't been able to figure out why I'm having this problem or how to reproduce it with a db that doesn't contain sensitive information, but I did notice that using `openKeyCursor` for the `key` and `keys` calls (in the IndexedDB driver) lets me work around the problem for now.

Is there a particular reason we're using `openCursor` instead of `openKeyCursor` for the `key` and `keys` call in the IndexedDB driver? Is there compatibility issues I'm not aware of? As far as I know, using `openKeyCursor` in this case would be much faster and is just more appropriate since we don't need any actual values for these calls.


lincolnthree:
@MeMark2 @tofumatt I've been testing this PR both locally, and now deployed to test, and yesterday to production. It seems this has settled some of the IDBExceptions that we've been seeing with Safari on iOS 13.

f:
Confirmed. This seems to be working.

tofumatt:
I figured it could have been a compatibility issue, but the two APIs don't seem to differ in their platform support. I'm reviewing this PR now (I was away last week), thanks for your patience! 🙂

tofumatt:
Just an update that this has been published on npm as is available in 1.7.4. Thanks again for the PR 😄

zhouxiaoping:
> Which versions of Edge does this happen with? Could you file an issue for this so we can address it specifically?
userAgent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/70.0.3538.102 Safari/537.36 Edge/18.18363

merge request for Edge18 compatibility：https://github.com/localForage/localForage/pull/971

PR #876, opened 2019-02-11
fix(setDriver): avoid overwriting self._driverSet on setDriver failure

sohkai:
On failure, the current behaviour overwrites any new `self._driverSet` that may be waiting to evaluate.


test/test.nodriver.js
```diff
@@ -81,4 +81,21 @@ describe('When No Drivers Are Available', function() {
             done();
         });
     });
+
+    it('allows a custom driver to be set immediately when no other drivers are found', function(done) {
```
sohkai: This test should always fail on the previous version

PR #909, opened 2019-10-01
Cleanup LICENSE

SheetJSDev:
Github does not currently recognize the license for this project.  [A project I help maintain](https://github.com/sheetjs/js-xlsx) had the same problem a long time ago, and it was explained that GH license check system looks for the full license text.  Merely including the notice blurb is insufficient.  This PR includes the full license text in a way that should be recognized by GH


SheetJSDev:
And just like that, it now says "Apache 2.0" in the bar :)

PR #913, opened 2019-11-15
[localForage]: Replaced PhantomJS with puppeteer

ossdev07:
Since, PhantomJS is an outdated project and no longer maintained. This patch replace PhantomJS with puppeteer.


thgreasi:
Thanks for opening this PR :+1: 
It seems that the test runner fails on travis with
```
Loading "Gruntfile.js" tasks...ERROR
>> SyntaxError: missing ) after argument list
```

Can you please take a look?

ossdev07:
ping @thgreasi, first time it got failed due to older node version. I updated the travis.yml file. Now only two tests are failing due to timeout. If we re-run the test, it will pass. 

So, can you please do the needful. Also, please do let me know if anything need to be done from my side.

ossdev07:
ping @tofumatt , I want to help you in migration from PhantomJS to Karma. Can you please involve me in this activity?

tofumatt:
Hi @ossdev07, sure thing. Right now progress on it is a bit slow, but I've made a note to try and get my existing branch pushed so it can be iterated on. I'll likely get to it sometime later in the week, so I'll ping you when it's ready and let you know what remains to do 😄

ossdev07:
ping @tofumatt, since localForage is a fast and simple storage library which is used by a lot of web app developers. It looks like the library doesn't support ARM architecture because of its dependency on PhantomJS(Not available for ARM and unmaintained) and I want to work on localForage library to give ARM support. So, can you please let know when your are planning to merge Karma related changes.

tofumatt:
Sorry about the delay! localForage should only have a devDependency for PhantomJS; the library itself is just JS and actually shouldn't have any npm dependencies.

I'll definitely get to this soon, sorry for the delay (been a bit busier than anticipated the past few weeks), but if you're experiencing issues running/building/installing localForage on ARM that would be great to file as a separate issue—I've not heard of that. I guess aside from doing development and running the tests. Could you clarify what you mean by "looks like the library doesn't support ARM architecture"?

ossdev07:
Basically I want to test this library on ARM architecture by issuing "npm install" followed by "npm test" command. Since, PhantomJS is not available for ARM and this is blocking me to test and verify. Can you please let me know if we can test this in another way? It will be a great help.

thgreasi:
My two bit on this @tofumatt @ossdev07 .
Imo this is a great start on moving away from the long-time-dead phantomjs and even though puppeteer main target is chrome, it will definitely improve the quality of our tests (since it's an actual browser, while phantomjs isn't).

TL&DR; I'm fine with it if we move to Puppeteer as an intermediate step before we move to karma, since I can only see benefits out of it.

PS: Puppeteer now has experimental support for Firefox as well
https://github.com/puppeteer/puppeteer#q-what-is-the-status-of-cross-browser-support

thgreasi:
And the tests do pass after a re-run :tada:

PR #932, opened 2020-02-14
Added async/await example

nainardev:
Come on, its 2020, async/await came in ES2017.
Easier to read
I suggest to remove Promises version as it is harder to read

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -60,6 +60,13 @@ localforage.getItem('somekey').then(function(value) {
     console.log(err);
 });
 
+// Async/await version:
+(async() => {
+    const value = await localforage.getItem('somekey');
+    // This code runs once the value has been loaded
+    // from the offline store.
+})()
+
 // Callback version:
 localforage.getItem('somekey', function(err, value) {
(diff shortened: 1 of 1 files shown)


docs/api.md
```diff
@@ -60,6 +60,13 @@ localforage.getItem('somekey').then(function(value) {
     console.log(err);
 });
 
+// Async/await version:
+(async() => {
```
tofumatt: Wrapping the example in an `async() => {}` function is _technically_ correct if this code is being written at the beginning of a file, but I think it's confusing and not very common from other `async`/`await` examples I've seen. It's also missing the error-handling of other example. Something more like:

```js
// Async/await version:
try {
    const value = await localforage.getItem('somekey');
    // This code runs once the value has been loaded
    // from the offline store.
    console.log(value);
} catch (err) {
    // This code runs if there were any errors.
    console.log(err);
}
```

would work better, I think, and also is in-line with other examples.

nainardev:
Sorry, for my behaviour 🙂. Will create a new PR with updated things

Issue #947, opened 2020-04-18
Typecheck of "storeName" during config values setting


himanshu199728:
Check of type of config value "store name" is needed as code may break if the user gives a number which through exception for replace() function.

PR #948, opened 2020-04-18
Check fix of storeName

himanshu199728:
Issue resolved of #947

### DIFF
--- a/src/localforage.js
+++ b/src/localforage.js
@@ -51,9 +51,9 @@ const DefaultConfig = {
 };
 
 function callWhenReady(localForageInstance, libraryMethod) {
-    localForageInstance[libraryMethod] = function() {
+    localForageInstance[libraryMethod] = function () {
         const _args = arguments;
-        return localForageInstance.ready().then(function() {
+        return localForageInstance.ready().then(function () {
             return localForageInstance[libraryMethod].apply(
                 localForageInstance,
                 _args
@@ -107,7 +107,7 @@ class LocalForage {
         this._dbInfo = null;
 
         this._wrapLibraryMethodsWithReady();
-        this.setDriver(this._config.driver).catch(() => {});
+        this.setDriver(this._config.driver).catch(() => { });
     }
 
     // Set any config values for localForage; can be called anytime before
@@ -128,8 +128,10 @@ class LocalForage {
             }
 
             for (let i in options) {
-                if (i === 'storeName') {
+                if (i === 'storeName' && typeof options[i] !== 'string') {
                     options[i] = options[i].replace(/\W/g, '_');
+                } else {
+                    return new Error('storeName must be a string');
                 }
 
                 if (i === 'version' && typeof options[i] !== 'number') {
@@ -153,15 +155,17 @@ class LocalForage {
         }
     }
 
+
+
     // Used to define a custom driver, shared across all instances of
     // localForage.
     defineDriver(driverObject, callback, errorCallback) {
-        const promise = new Promise(function(resolve, reject) {
+        const promise = new Promise(function (resolve, reject) {
             try {
                 const driverName = driverObject._driver;
                 const complianceError = new Error(
                     'Custom driver not compliant; see ' +
-                        'https://mozilla.github.io/localForage/#definedriver'
+                    'https://mozilla.github.io/localForage/#definedriver'
                 );
 
                 // A driver name should be defined and not overlap with the
@@ -190,9 +194,9 @@ class LocalForage {
                     }
                 }
 
-                const configureMissingMethods = function() {
-                    const methodNotImplementedFactory = function(methodName) {
-                        return function() {
+                const configureMissingMethods = function () {
+                    const methodNotImplementedFactory = function (methodName) {
+                        return function () {
                             const error = new Error(
                                 `Method ${methodName} is not implemented by the current driver`
                             );
@@ -223,7 +227,7 @@ class LocalForage {
 
                 configureMissingMethods();
 
-                const setDriverSupport = function(support) {
+                const setDriverSupport = function (support) {
                     if (DefinedDrivers[driverName]) {
                         console.info(
                             `Redefining LocalForage driver: ${driverName}`
@@ -314,7 +318,7 @@ class LocalForage {
         }
 
         function initDriver(supportedDrivers) {
-            return function() {
+            return function () {
                 let currentDriverIndex = 0;
 
                 function driverPromiseLoop() {
--- a/src/localforage.js
+++ b/src/localforage.js
@@ -118,40 +118,44 @@ class LocalForage {
         // If the options argument is an object, we use it to set values.
         // Otherwise, we return either a specified config value or all
         // config values.
-        if (typeof options === 'object') {
-            // If localforage is ready and fully initialized, we can't set
-            // any new configuration values. Instead, we return an error.
-            if (this._ready) {
-                return new Error(
-                    "Can't call config() after localforage " + 'has been used.'
-                );
-            }
-
-            for (let i in options) {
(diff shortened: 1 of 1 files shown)


src/localforage.js
```diff
@@ -118,50 +118,58 @@ class LocalForage {
...
+            } else {
+                return this._config;
             }
-
-            return true;
-        } else if (typeof options === 'string') {
-            return this._config[options];
-        } else {
-            return this._config;
+        } catch (err) {
```
tofumatt, line 157: We shouldn't catch any error here; instead this should throw as it did previously.

src/localforage.js
```diff
@@ -118,50 +118,58 @@ class LocalForage {
...
+                    );
                 }
 
-                if (i === 'version' && typeof options[i] !== 'number') {
-                    return new Error('Database version must be a number.');
+                for (let i in options) {
+                    if (i === 'storeName' && typeof options[i] !== 'string') {
+                        options[i] = options[i].replace(/\W/g, '_');
+                    } else {
+                        return new Error('storeName must be a string');
```
tofumatt, line 135: This is a helpful touch, and I guess it isn't a breaking change since `.replace` on anything that isn't a `String` would fail anyway (since it isn't a function).

src/localforage.js
```diff
@@ -118,50 +118,58 @@ class LocalForage {
...
+                if (this._ready) {
+                    return new Error(
+                        "Can't call config() after localforage " + 'has been used.'
+                    );
                 }
 
-                if (i === 'version' && typeof options[i] !== 'number') {
-                    return new Error('Database version must be a number.');
+                for (let i in options) {
+                    if (i === 'storeName' && typeof options[i] !== 'string') {
```
tofumatt, line 132: Reading this code it seems like this `if` check will fail (calling the `else` block below) for any option key that isn't `storeName`. You'll want to nest the check for `(typeof options[i] !== 'string')` inside this `if` statement or this will break other things.

src/localforage.js
```diff
@@ -118,50 +118,58 @@ class LocalForage {
...
-            if (this._ready) {
-                return new Error(
-                    "Can't call config() after localforage " + 'has been used.'
-                );
-            }
-
-            for (let i in options) {
-                if (i === 'storeName') {
-                    options[i] = options[i].replace(/\W/g, '_');
+        try {
```
tofumatt, line 121: I don't think this `try/catch` block here is helpful to add, as it will prevent unexpected errors from throwing.

src/localforage.js
```diff
@@ -314,7 +322,7 @@ class LocalForage {
         }
 
         function initDriver(supportedDrivers) {
-            return function() {
+            return function () {
```
tofumatt, line 325: Not sure why these code changes were made, but they're failing the linter; could you please revert these style/whitespace changes?

himanshu199728: Got this due to correcting the indentation by some extension. I'll revert it back.
Can I use assert library to write test cases in this project??

himanshu199728: I think test cases in the test folder needed to be more organized in categories for manging and ease to find.

himanshu199728: can you check it again

himanshu199728: I didn't get any update from you, sir.

package.json
```diff
@@ -69,6 +69,7 @@
     "url": "http://github.com/localForage/localForage/issues"
   },
   "dependencies": {
+    "jslint": "^0.12.1",
```
tofumatt, line 72: This should be in `devDependencies`, not `dependencies`, as you don't need `jslint` to run the repo.

That said: this shouldn't need to be here at all. If we're using a linter it should be `eslint`, but we aren't using a linter (or, specifically, `jslint`) anywhere in the codebase or this PR. Could you remove this line please?

test/test.config.js
```diff
@@ -269,4 +269,16 @@ describe('Config API', function() {
         expect(configResult.toString()).to.be(error);
         done();
     });
+
+    it('should throw error on config.storeName not string', function(done) {
```
tofumatt, line 273: Just a wording suggestion here to clean up the test:

```suggestion
    it('should throw error if config.storeName is not a string', function() {
```

(You also don't need to use `done()` here since this isn't an asynchronous test.)

test/test.config.js
```diff
@@ -269,4 +269,16 @@ describe('Config API', function() {
...
+        let configResult = localforage.config({
```
tofumatt, line 274:
```suggestion
        const configResult = localforage.config({
```

test/test.config.js
```diff
@@ -269,4 +269,16 @@ describe('Config API', function() {
...
+            description: 'The offline datastore for my cool app',
+            driver: localforage.driver(),
+            name: 'My Cool App',
+            storeName: 123,
+            version: 2.0
+        });
+        expect(configResult).to.be(new Error('storeName must be a string'));
+        done();
```
tofumatt, line 282:
```suggestion
```

package.json
himanshu199728: Ok

himanshu199728: Sorry, I forget that this type of dependency came under dev dependencies.

test/test.config.js
himanshu199728: Ok

test/test.config.js
himanshu199728: Docs also need to be updated for the same right in the readme. Right??

himanshu199728: Sorry, my branch name is not following the remaining branches convention of writing using - instead of _. I will take care next time.

tofumatt: No worries! I think the docs are okay, `storeName` says:

> The name of the datastore. In IndexedDB this is the dataStore, in WebSQL this is the name of the key/value table in the database. Must be alphanumeric, with underscores. Any non-alphanumeric characters will be converted to underscores.

If you wanted to add that it must be a `String` though, that's fine 🙂

himanshu199728: Does this branch is ready to merge into master

himanshu199728: Why there is no dev branch in this project and only master + feature branches. Won't it be complex to handle the project in such a way??

tofumatt: Nope, this approach has worked just fine for us. 🙂

I'm not a big fan of `develop` or `dev` branches personally, and for this project it would be an extra maintenance burden.

himanshu199728: > Does this branch is ready to merge into master?

??

himanshu199728: > Nope, this approach has worked just fine for us. 🙂
> 
> I'm not a big fan of `develop` or `dev` branches personally, and for this project it would be an extra maintenance burden.

Ok. Cool.

package.json
```diff
@@ -23,6 +23,7 @@
     "test": "node -e \"require('grunt').cli()\" null test"
   },
   "devDependencies": {
+    "jslint": "^0.12.1",
```
tofumatt, line 26: Where is the code using `jslint`? I mentioned tooling like this belongs in `devDependencies` if used, but I don't see it being used anywhere in this PR so I think it's best to remove this.

test/test.config.js
tofumatt: > Does this branch is ready to merge into master?

We merge PRs into the `master` branch, yes.

package.json
himanshu199728: Ok

himanshu199728: Done

himanshu199728: Is it now ready to merge??

tofumatt: Please be patient; this is an open-source project and we don't always have the time to respond to PRs right away. I do check my issue queues/notifications but I can't get to things right away 🙂

himanshu199728: Ok. No worry.

Issue #952, opened 2020-05-01
set key-value pair in bulk should be provided in localstorage driver(database)


himanshu199728:
Currently, we are storing key-value pair data and calling for each time each if we want to set multiple key-value pairs at a time.

**Currently**: setItem(<key>, <value>);

**My Expectation**: setItems([{key:<key>, value: <value>}, [key :<key>, value :<value>]...]);

Sometimes we require to cache more than one DB connection details to store or multiple photos data as blob store in bulk in storage.

thgreasi:
Hi,
Have you checked https://github.com/localForage/localForage-setItems ?
It sounds like it does exactly what you need for all supported drivers.
We decided to not make this part of the core localforage api, at least initially, since there was also a different approach in the works at that point.

I will close this for now, since I think that the extension covers this use case, but feel free to re-open.

himanshu199728:
#953. Can we look into this? or this is a useless change.

PR #954, opened 2020-05-03
setItems in bulk in localStorage

himanshu199728:
#952, #953

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,3 +38,4 @@ doc/
 
 # typescript cache
 .tscache
+.vscode/*
--- a/src/drivers/indexeddb.js
+++ b/src/drivers/indexeddb.js
@@ -6,6 +6,7 @@ import executeCallback from '../utils/executeCallback';
 import executeTwoCallbacks from '../utils/executeTwoCallbacks';
 import normalizeKey from '../utils/normalizeKey';
 import getCallback from '../utils/getCallback';
+import DATA_ERROR from '../msg/data.error';
 
 // Some code originally from async_storage.js in
 // [Gaia](https://github.com/mozilla-b2g/gaia).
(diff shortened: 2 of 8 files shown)


himanshu199728:
I'm not getting this why the build failed.

thgreasi:
Repeating what I mentioned in #952
> Hi,
Have you checked https://github.com/localForage/localForage-setItems ?
It sounds like it does exactly what you need for all supported drivers.
We decided to not make this part of the core localforage api, at least initially, since there was also a different approach in the works at that point.

Should we close this?

himanshu199728:
> Repeating what I mentioned in #952
> 
> > Hi,
> > Have you checked https://github.com/localForage/localForage-setItems ?
> > It sounds like it does exactly what you need for all supported drivers.
> > We decided to not make this part of the core localforage api, at least initially, since there was also a different approach in the works at that point.
> 
> Should we close this?

Ok. I have closed the pull request regarding this. Will start working on some other issue.

thgreasi:
Cool, thanks :+1:

himanshu199728:
Actually in code I was trying to have setItems api for localStorage driver and if user try to call setItems() for other drivers then he will get "unsupported service" response in Promise as return.

thgreasi:
Indeed, but localforage-setitems also works for IDB & Websql, as well as any possible custom driver used (like localforage-cordova-sqlite-driver). Are you suggesting something different?

himanshu199728:
I'm saying in context of version of this library setItems() will be available for localStorage but later it will be available for other drivers also. We will have setItems() available for all supported driver but excluding localStorage driver all will return new Promise((resolve, rejext)=>{resolve("Service not available")}). Like if in some authentication service we only having user-password authentication available and if the developer tries to make it for OTP(whose development is not done) it will return "Service not available" as a response.

PR #964, opened 2020-05-26
Fix #856: Be more lenient detecting window.fetch, to support polyfills.

### LINKED ISSUE #856: localforage driver should default to indexeddb, instead causing race conditions with WebSQL
Interesting bug that took me forever to track down.

Basically, unless you explicitly set driver in localforage.config, Chrome and Safari are indeterministic about which driver they will use, resulting in what "seems" like data loss. This happens at random.

So basically, I have 2 databases, and I'll be connected to one of them through localforage on any given reload.

lincolnthree:
Fixes https://github.com/localForage/localForage/issues/856

### DIFF
--- a/src/utils/isIndexedDBValid.js
+++ b/src/utils/isIndexedDBValid.js
@@ -17,9 +17,7 @@ function isIndexedDBValid() {
             !/Chrome/.test(navigator.userAgent) &&
             !/BlackBerry/.test(navigator.platform);
 
-        var hasFetch =
-            typeof fetch === 'function' &&
-            fetch.toString().indexOf('[native code') !== -1;
+        var hasFetch = typeof fetch === 'function';
 
         // Safari <10.1 does not meet our requirements for IDB support
         // (see: https://github.com/pouchdb/pouchdb/issues/5572).


src/utils/isIndexedDBValid.js
```diff
@@ -17,9 +17,7 @@ function isIndexedDBValid() {
             !/Chrome/.test(navigator.userAgent) &&
             !/BlackBerry/.test(navigator.platform);
 
-        var hasFetch =
-            typeof fetch === 'function' &&
-            fetch.toString().indexOf('[native code') !== -1;
+        var hasFetch = typeof fetch === 'function';
```
mareksuscak, line 20: I'm a bit worried that the entire point of checking for a native implementation was to ensure that the script's running in a browser with a native fetch because that's an indication that the given browser supports IndexedDb. This is also known as feature inference. 

After we make this change, polyfilling fetch would trick localForage into thinking that IndexedDb is also supported which might not be the case.

tofumatt: That's quite true, though that code is also a reflection of the state of JS at the time.

Out-of-scope of this PR, I think it would benefit _future_ versions of localForage to take a more "evergreen browser" support approach and leave the 1.x branch to maintain broader compatibility. The original two selling points for localForage were compatibility and ease-of-use, but the former is much less an issue over five years after its introduction. Effectively, browsers support IndexedDB and localForage exists to work around quirks and provide a nicer API.

I think we should make this change, but I'm thinking it might be best released in a new major version.

I've been trying to focus on localForage 2.0 for some time, but other things have consistently interrupted 😅. I think I should actually have some time for it this month, so I'll see what I can do. 🙂

mareksuscak: As an alternative workaround, we could tackle this problem similar to what [jQuery.noConflict](https://api.jquery.com/jquery.noconflict/) did. 

We could write and document a tiny JS snippet that the user of this library would need to insert before the polyfill that would set a global variable on the window and we could use the global variable in this condition.

Thoughts?

tofumatt: My first reaction is that it sounds a bit complicated to explain, and could be a bit frustrating to implement as well. Of course, the tradeoff is it not working as-expected with a modified `fetch`, so fair enough. I'd, of course, prefer libraries weren't modifying built-in `fetch`, but I get why it might happen 😄

I'll think about the best way to address this and try to sort out timelines for breaking changes. I'd really like to start improving the internals of localForage and modernise the library, including removing much of the older compatibility hacks/fixes and the callback support for modern JS environments. That'd be the perfect time to introduce this fix, in my opinion, so if it isn't too late it's what I'd prefer.

But if that doesn't happen soon, then we should explore shipping this fix in a non-breaking way, even if it involves a type of `noConflict` route.

mareksuscak: > I'd, of course, prefer libraries weren't modifying built-in fetch, but I get why it might happen.

Agreed. In our case, it's Fullstory and Sentry that both wrap the native fetch so that they can collect breadcrumbs.

mareksuscak:
BTW I just came across this PR https://github.com/localForage/localForage/pull/799 which appears to have fixed this issue in the past. This makes me wonder if and/or why it got rolled back?

mareksuscak:
Any plans to tie this PR off?

lincolnthree:
Bump? Sorry to be a pain. This is still a blocking issue for us.

src/utils/isIndexedDBValid.js
jacobg: I'm also running into this issue. Bugsnag.js SDK polyfills fetch() for logging breadcrumbs. When emulating iOS on desktop Chrome browser, it switches to WebSQL. The result is that it uses a different database than when either not emulating or when emulating Android. It would be great if there were a config option to bypass the fetch() check, or to just force a particular driver. It looks like the source code hides all possible monkey patches inside closures, and I'd rather not fork this repo. And the config API rejects specifying drivers that are not "supported".

jacobg:
I'm also running into this issue. Bugsnag.js SDK polyfills fetch() for logging breadcrumbs. When emulating iOS on desktop Chrome browser, it switches to WebSQL. The result is that it uses a different database than when either not emulating or when emulating Android. It would be great if there were a config option to bypass the fetch() check, or to just force a particular driver. It looks like the source code hides all possible monkey patches inside closures, and I'd rather not fork this repo. And the config API rejects specifying drivers that are not "supported".

PR #972, opened 2020-06-24
api.md update - Multiple stores in same instance

owenfar:
I kept it simple. Let me know if I should add/change something or do so as you see fit :) Thank you for listening and for the quick response.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -553,7 +553,6 @@ var tableOne = localforage.createInstance({
 
 // Create table 2 in databaseName
 var tableTwo = localforage.createInstance({
-    driver      : localforage.INDEXEDDB,
     name        : dbName,
     storeName   : 'tableTwo',
     description : '...'
--- a/docs/api.md
+++ b/docs/api.md
@@ -545,7 +545,6 @@ const dbName = 'databaseName';
 
 // Create table 1 in databaseName
 var tableOne = localforage.createInstance({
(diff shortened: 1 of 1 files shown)


docs/api.md
```diff
@@ -537,3 +537,27 @@ localforage.dropInstance({
...
+var tableOne = localforage.createInstance({
+    driver      : localforage.INDEXEDDB,
+    name        : dbName,
+    storeName   : 'tableOne',
+    description : '...'
+});
+
+// Create table 2 in databaseName
+var tableTwo = localforage.createInstance({
+    driver      : localforage.INDEXEDDB,
```
tofumatt, line 556: Technically this isn't needed, as in WebSQL this will use a table and in localStorage it will namespace the keys.
```suggestion
```

docs/api.md
```diff
@@ -537,3 +537,27 @@ localforage.dropInstance({
...
 When invoked with an object specifying only a `name` property, it drops the specified "database" (and all its stores).
+
+# Multiple Stores
+
+```js
+const dbName = 'databaseName';
+
+// Create table 1 in databaseName
+var tableOne = localforage.createInstance({
+    driver      : localforage.INDEXEDDB,
```
tofumatt, line 548: See my other comment, technically we can omit this and should 😄 
```suggestion
```

owenfar:
Thank you @tofumatt for all the time and effort you put into localForage :)

Issue #940, opened 2020-03-12
Release in ESM format


nainardev:
>No web-optimized "module" entrypoint was found in its package.json manifest"

![Screenshot_20200313-000704](https://user-images.githubusercontent.com/59839368/76556046-d34faf00-64be-11ea-9371-48fc346a2f83.png)

quentincaffeino:
Wouldn't be enough adding this line to package json?
`"module": "src/localforage.js",`
I tried it and it worked.

tofumatt:
Works for me! Could you open a PR? I'll happily approve it and ship a new release with it today 😄

tofumatt:
1.8.0 has been published and is available on NPM. Thanks! https://www.npmjs.com/package/localforage

Samuel-Therrien-Beslogic:
Since https://github.com/localForage/localForage/pull/977 has been reverted, this should likely be re-opened.
IE11 is also finally dead

emeryao:
+1
any updates here?

sysmat:
any updates?

PR #977, opened 2020-07-19
Added module to package.json for esm suport

quentincaffeino:
Closes: #940

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,6 +63,7 @@
     "webpack-dev-server": "^1.10.1"
   },
   "main": "dist/localforage.js",
+  "module": "src/localforage.js",
   "typings": "typings/localforage.d.ts",
   "bugs": {
     "url": "http://github.com/localForage/localForage/issues"


tofumatt:
Thanks so much! Sorry I didn't get to this yesterday, but I'm shipping a new version now! 😄

klarstrup:
This is probably going to break IE11 etc. that don't support all the newness used in `src/localforage.js`

quentincaffeino:
How is this able to break anything? Just use precompiled option as you've used before. This is just an option for newer modular bundles that support esm.
Correct me if I'm wrong.

klarstrup:
Webpack(and probably rollup) always uses `module` if provided, and most consumers are not set up to transpile code from node_modules so it gets included in it's raw modern form.

quentincaffeino:
True. Can't say anything about webpack, but from what I've found rollup resolves by default to module, and uses main as a fallback ([Docs](https://github.com/rollup/plugins/tree/master/packages/node-resolve#mainfields), [Code](https://github.com/rollup/plugins/blob/master/packages/node-resolve/src/util.js#L31...L45)).

tofumatt:
Yup, this seems to have broke people's builds. I'm reverting it now, sorry.

rozek:
since localForage does not seem to become available as an ECMAScript module in the near future, [I just copied the source code and bundled it myself into an ESM](https://github.com/rozek/localforage-esm) - now I can import with ease, even directly from within a browser (i.e., without requiring a build environment)

PR #980, opened 2020-07-24
Update localforage.d.ts

Glandos:
getItem can return null when the item doesn't exist

### DIFF
--- a/typings/localforage.d.ts
+++ b/typings/localforage.d.ts
@@ -15,7 +15,7 @@ interface LocalForageOptions extends LocalForageDbInstanceOptions {
 }
 
 interface LocalForageDbMethodsCore {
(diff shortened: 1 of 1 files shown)


Glandos:
It's a nice debatable question. I personally opened this because I started using localforage, and the bad Typescript signature leads me to a bug, with an uncatched null value.
In my opinion, this was a bug that should be resolved. If it breaks build, it means that projects had bugs that should be resolved too. However, it also requires a huge warning sign in the changelog :wink: 

But if the 2.0 rewrite is not far from complete (let's say, 6 months), then it doesn't really make sense to annoy people with that. This existing issue should guide people to a workaround in the meantime.

tofumatt:
Fair enough, let's ship this as 1.9.

I'll put it together tomorrow 😄

tofumatt:
Just an update: this was published as localForage 1.9 today. Thanks!

PR #983, opened 2020-08-09
Export interfaces

whitecolor:
Following my comment made four years ago =) https://github.com/localForage/localForage/issues/600#issuecomment-243891174

Export interfaces for convenient driver creation.

### DIFF
--- a/typings/localforage.d.ts
+++ b/typings/localforage.d.ts
@@ -4,7 +4,7 @@ interface LocalForageDbInstanceOptions {
     storeName?: string;
 }
 
-interface LocalForageOptions extends LocalForageDbInstanceOptions {
+export interface LocalForageOptions extends LocalForageDbInstanceOptions {
     driver?: string | string[];
 
     size?: number;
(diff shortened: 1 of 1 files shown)


thgreasi:
The tests seems to not pass.
I think that b/c we are using a module definition (b/c of the UMD export we do), you can't use `export` at the top level and you need to define the interfaces inside the `module` definition.

whitecolor:
Actually I just noticed that interfaces are available without the need to explicitly import. Which is not nice, but at least it works.

whitecolor:
@thgreasi
though there is an issue with this

whitecolor:
thi

thgreasi:
> Because it allows to use modules even if not installed in (available for) particular package for example in the workspace project

@whitecolor can you clarify what you mean with this a bit more? The *installed* part is what confused me.
Do you mean that in a project that you have LocalForage installed, you can use the interfaces even w/o importing or `/// reference`-ing them?

whitecolor:
> Do you mean that in a project that you have LocalForage installed, you can use the interfaces even w/o importing or /// reference-ing them?

I think everything is ok.

So one still have to `import from 'localforage'`, they are not globally visible.

```ts
import type localforage from 'localforage'

const options: LocalForageOptions 
```

thgreasi:
Can you also try fixing the tests?
Otherwise I'm afraid we can't merge this.

thgreasi:
Let me know if you would prefer me or someone else to pick this up.

whitecolor:
> Let me know if you would prefer me or someone else to pick this up.

Yes, I don't know to fix the tests. As I said actually I'm ok with the current situation, exporting interface seems not actually necessary. But probably it would more correct to make them only available via explicit importing.

Issue #673, opened 2017-03-15
Support ES6 Maps & Sets in serializer (thus in WebSQL and localStorage)


finnef:
When saving native ES6 Maps or Sets to a WebSQL or localStorage backend the serialiser is used to convert the value. This serialiser checks for binary objects, but does a JSON.stringify() on other objects. Maps and Sets are cast to empty objects this way. 
You could add detection for these object types and save them converted. See 'Arbitrary Maps as JSON via Arrays of pairs' in http://www.2ality.com/2015/08/es6-map-json.html.

You could use the same mechanism as for blobs and type_arrays

PR #989, opened 2020-08-11
updated to show that Map and Set are now supported

therealadityashankar:
they've been supported since #673 so I think it should be reflected in the docs


therealadityashankar:
docs updated led to a travis-ci check failure, not really my fault tbh

PR #992, opened 2020-09-14
Update api.md

ali-master:
Minor improvement in API doc

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -51,6 +51,7 @@ These APIs deal with getting and setting data in the offline store.
 ## getItem
 
 ```js
+// Promise version
 localforage.getItem('somekey').then(function(value) {
     // This code runs once the value has been loaded
     // from the offline store.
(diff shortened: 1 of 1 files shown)


docs/api.md
```diff
@@ -67,9 +68,7 @@ localforage.getItem('somekey', function(err, value) {
     console.log(value);
 });
 
-Or, use `async`/`await`:
-
-```js
+// Or, use `async`/`await`:
```
tofumatt, line 71: I'd rather these not be code comments but text, as they're instructions to the reader.

tofumatt (changes requested):
The instructions to use `async`/`await` aren't really appropriate as code comments. They aren't describing the code: they're instructing the developer/reader directly. I'm fine to annotate the promises version a bit more clearly, but let's revert the async/await change 😄

ali-master:
Thanks for replying.

If you copy the Async/Await API doc with ```Or, use `async`/`await`:``` line, Javascript will throw a syntax error in this line, So It would be better if we comment this line to prevent getting an error.

I saw that mistake when our colleagues copied those lines to use. But finally, if you think it's not so good, I will revert.

ali-master:
Any news?

tofumatt:
As those aren't code comments, I'd prefer they be outside the code. Generally only code inside code blocks should be treated as being able to be copied 😄

ali-master:
I've uncommented that line. Check it again, please

Issue #910, opened 2019-10-08
storage size


sopov:
Hello,

How to properly control the size of the storage?

I have few instances of localforage,
when ~ 3mb of data (166 keys) stored I try call `setItem` for new key I have in developer console error.

`InvalidStateError: Failed to read the 'error' property from 'IDBRequest': The request has not finished.`

after this error other instance is available (exactly for reading, not sure for saving), butершы instance (with error) not answer more (incl getItem), but before setItem I can request any data (getItem, keys, etc).

if I call like
`localforage.setItem(..).then(console.log).catch(console.error) `
catch -- never called

How do I write it right?
```
localforage.setItem(..).catch(
    function () {
        localforage.clean().then(function () {
            localforage.setItem(..)
        })
    }
)
```

* Safari 13.0.1

sopov:
I can not `clear` and have same issue
````
localforage.keys().then(function (keys) {
    console.log('KEYS: ' + keys.length)
    if (keys.length > 100) {
        console.log('CLEAR')
        localforage.clear().then(
            function () {
                console.log('DONE PAGES')
            }
        )
    } else {
        console.log('ALL IS OK')
    }
})
```

![CleanShot 2019-10-08 at 08 18 33](https://user-images.githubusercontent.com/325038/66369236-3f743700-e9a4-11e9-9c2f-10038a098422.png)

sopov:
I don't think it's a size of storage, maybe it's some kind of data that's stored in DB.
I try reproduce issue again, some time with less data I have same issue, now I have more 600keys & 10mb -- no issue.

carlwoodward:
Hey, is there any chance this the new version of Safari? I'm hitting this on iOS Safari and Desktop and I've tried multiple IndexedDB libraries and rolling my own. My test system has syncs the same data between these browsers. Any thoughts?

galfert:
Safari on iOS 13 (and 12) indeed has some serious problems with IndexedDB:

https://bugs.webkit.org/show_bug.cgi?id=197050
https://bugs.webkit.org/show_bug.cgi?id=202705

johnozbay:
I just got hit by this as well. 
My findings are also in line with @sopov's. 

+ It's not related to the size. 

+ It seems like it could be related to the content of the item being written, because some `setItem` calls work some fail.

+ If `setItem` returns `InvalidStateError: Failed to read the 'error' property from 'IDBRequest': The request has not finished.` when you try to remove the same item, `removeItem` returns the same error.

[...]

PR #1000, opened 2020-11-05
fix: GH-910 separate error handlers

tomsaleeba:
This isn't a complete fix for https://github.com/localForage/localForage/issues/910 but it at least makes it so
 1. the original error isn't masked
 1. the promise is rejected so calling code can handle the error

I figured the ternary expression was only there because there was a single handler function.

Once I made this change, I could see that I was getting a transaction abort not an error. The distinction was hard to tell before just based on line numbers as they used the same function. The error for the aborted transaction is:
image

I don't know what's causing that, but that's not relevant to this PR. Although I'll happily accept any guesses on where to look for it.

CC @mimittqq who [requested this](https://github.com/localForage/localForage/issues/910#issuecomment-576562369) in the linked issue too.

CC @tokmakoff


tomsaleeba:
Actually, swapping PR to draft. I should've looked closer at my screenshot. I'm not catching that error after all.

I'll figure it out.

